PR #6, opened 2018-01-30
Uses HTTP GET for type Query queries

tlgimenes:
This feature forces the use of GET for type Query types, while still using POST for mutations


react/internal/http-switch-link.js
```diff
@@ -0,0 +1,56 @@
+/*
+ * This file defines a linker middleware used for changing http verbs, allowing
+ * cache in CDNs and other http-based caching content delivery networks
+ *
+ * If the query type is Query, the request will be made using an HTTP GET
+ * If the query type is Mutation, the request will be made using HTTP POST
+ *
+ */
+import {ApolloLink} from 'apollo-link'
+import { prop, map, contains, isEmpty, isNil } from 'ramda'
```
firstdoit: We shouldn't use ramda in the runtime as it will add to bundle size. Incidentally, you were able to use it without declaring because we have a node modules leak in the builder hub. Need to look into that.

react/internal/http-switch-link.js
```diff
@@ -0,0 +1,56 @@
...
+import { parse, stringify } from 'querystring'
```
firstdoit: Mind the spaces between curly braces.

react/internal/http-switch-link.js
```diff
@@ -0,0 +1,56 @@
...
+import { print } from 'graphql'
+
+function removeUnusedFields(obj) {
+  Object.keys(obj).forEach(key => (isNil(obj[key]) || isEmpty(obj[key])) && delete obj[key])
+
+  return obj
+}
+
+export const createHttpSwitchLink = (uri) => {
+  const [baseUrl, querystring] = uri.split('?')
```
firstdoit: Splitting on ? is not a safe way to find a query. You should parse the url. Not sure what's the best way in the browser, there is probably something native

react/internal/http-switch-link.js
```diff
@@ -0,0 +1,56 @@
...
+  let queryObj = (querystring) ? parse(querystring) : {}
+
+  return new ApolloLink((operation, forward) => {
+    let targetUri = uri
+    const {query, variables, operationName} = operation
+    const {fetchOptions = {}, http: httpOptions = {}} = operation.getContext()
+    const method = contains('mutation',map(prop('operation'), prop('definitions')(query))) ?
```
firstdoit: We will drop the ramda but this could be a pluck

react/internal/http-switch-link.js
tlgimenes: Ramda removed. Had to do some ugly code though :(

react/internal/http-switch-link.js
tlgimenes: sorry, I have to fix my lint

react/internal/http-switch-link.js
tlgimenes: used package `url` for this. It actually helped making the code cleaner

react/internal/http-switch-link.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * If the query type is Query, the request will be made using an HTTP GET
+ * If the query type is Mutation, the request will be made using HTTP POST
+ *
+ */
+import {ApolloLink} from 'apollo-link'
+import {parse, format} from 'url'
+import {print} from 'graphql'
+
+const isEmpty = (val) => (val === undefined || val === "" || val === {} || val === [])
```
firstdoit: `val === {}` will always evaluate to false because `{}` declares a new object and `===` checks for referential equality.

firstdoit: Same for array.

firstdoit: (Take a look at the reference values section here: http://firstdoit.github.io/js-closures-workshop/)

mairatma:
It seems like there was some problem with the rebase you made in this branch @tlgimenes. Some of @firstdoit's commits from master are showing up in between old and new commits in this branch. Can you fix this? The pr should only have your commits for this.

react/internal/http-switch-link.js
```diff
@@ -7,42 +7,51 @@
...
-    let targetUri = uri
+    const targetUri = Object.assign({}, parsedUri)
     const {query, variables, operationName} = operation
     const {fetchOptions = {}, http: httpOptions = {}} = operation.getContext()
-    const method = contains('mutation',map(prop('operation'), prop('definitions')(query))) ?
-      'POST' :
-      'GET'
+    const method = hasMutationField(query) ? 'POST' : 'GET'
 
     if(method === 'GET') {
```
mairatma: Space missing after `if`.

react/internal/http-switch-link.js
```diff
@@ -7,42 +7,51 @@
...
-import { prop, map, contains, isEmpty, isNil } from 'ramda'
-import { parse, stringify } from 'querystring'
-import { print } from 'graphql'
+import {parse, format} from 'url'
+import {print} from 'graphql'
+
+const isEmpty = (val) => (val === undefined || val === "" || val === {} || val === [])
+const isNil = (val) => (val === undefined || val === null)
 
 function removeUnusedFields(obj) {
```
mairatma: This same function is in three code bases now (render-server, graphql-server and render-runtime), is it really necessary? Optimizations are only worth it if they cover cases that happen a lot. Can you explain the use case you're trying to optimize for here?

react/internal/http-switch-link.js
```diff
@@ -7,42 +7,51 @@
...
+    const method = hasMutationField(query) ? 'POST' : 'GET'
 
     if(method === 'GET') {
       fetchOptions['method'] = method
 
-      const fullQueryObj = removeUnusedFields({
-        ...queryObj,
+      targetUri.query = removeUnusedFields({
+        ...targetUri.query,
         'query': print(query).replace(/\s\s+/g, ' '),
```
mairatma: We don't use quotes for object keys that don't need them, so this should just be
```js
query: print(query).replace(/\s\s+/g, ' '),
```

react/internal/http-switch-link.js
```diff
@@ -7,42 +7,51 @@
...
         operationName
       })
 
-      targetUri = baseUrl.concat(`?${stringify(fullQueryObj)}`)
+      // need to delete in order for format function to work parsing the query instead of search field
+      delete targetUri.search
     }
 
-    operation.setContext({uri: targetUri, fetchOptions})
+    operation.setContext({uri: format(targetUri), fetchOptions})
```
mairatma: It's best to pass the old context here together with this new information, as otherwise some data may be lost (`setContext` replaces the last context, as I found out when using it inside `graphql-server`, it doesn't just add the new data to it).

So in this case, something like:

```js
operation.setContext({
  ...operation.getContext(),
  uri: format(targetUri), fetchOptions
})
```

react/internal/http-switch-link.js
```diff
@@ -30,11 +29,13 @@ const hasMutationField = (queryTree) => {
 
 export const createHttpSwitchLink = (uri) => {
   const parsedUri = parse(uri, true)
+  // delete search in order to format to work as expected
+  parsedUri.search && delete parsedUri.search
```
mairatma: Can you explain the use case where this is important? That's not clear to me.

By the way, comments are most useful to explain things such as this: why the code is there, instead of just what it's doing (since the latter can be deduced by reading the code, but the former not always can).

tlgimenes: There is a bug (or a feature) in the url parser package that we are using that, if the field `search` is present in the object, the formatter won't use the `query` field for generating the URI. Since we want to force the formatter to use the `query` field, we need to drop the `search` field

mairatma: But then we'll just lose the information in this `search` field? What is it usually used for? Can we keep it in some other way?

PR #9, opened 2018-02-03
[WIP] Refactor runtime to use context correctly and other improvements

firstdoit:
- [x] Use static class properties when possible
- [x] No more messy globals running around
- [x] No more mutating global state unnecessarily
- [x] Each nested extension point gets it's `params` via props
- [x] Implement componentDidCatch on ExtensionPoint
- [x] Use `eventemitter3` and reference it through context
- [x] Assets are loaded on demand when extension points render (support for async extension points!)
- [x] Update subscriptions if component changes
- [x] Mount ApolloProvider inside RenderProvider
- [x] Pass prefetchPage in context


firstdoit:
Taking the chance to do the pages-graphql migration too, so I'll need to test this further. WIP again

firstdoit:
This PR changed beyond recognition, dropping it for one that makes more sense

PR #11, opened 2018-02-06
Use external locale data

brunoabreu:


public/externals.json
```diff
@@ -26,6 +26,9 @@
     },
     {
       "path": "https://unpkg.com/vtex-tachyons@1.0.2/tachyons.css"
+    },
+    {
+      "path": "https://unpkg.com/intl@1.2.5/dist/Intl.js"
```
firstdoit: Shouldn't we add this do the dynamic poly in render? Recent browsers have it

react/components/Router.js
```diff
@@ -9,7 +9,7 @@ import ExtensionPoint from '../ExtensionPoint'
 import IntrospectionFetcher from './IntrospectionFetcher'
 
 function isRelative(path) {
-  return path[0] !== '/'
+  return !path.startsWith('http://') && !path.startsWith('https://')
```
firstdoit: Why was this change necessary?

public/externals.json
brunoabreu: OK, but at least it needs to be loaded for SSR, because nodejs comes only with english locales.

react/components/Router.js
brunoabreu: Because `isRelative('https://unpkg.com/react-intl@2.4.0/locale-data/en.js')` being true doesn't make sense.

But I was looking where this function was being used and I think this is legacy code:

```js
function prefix(path) {
  const prefix = isRelative(path) ? '/assets/' : ''
  return `${prefix}${path}`
}
```

We don't even have this `/assets/*` route in render-server anymore.

I'll remove this code.

PR #14, opened 2018-02-09
Feature/persisted queries

tlgimenes:
Adds link to enable persisted queries.

The runtime always tries to send the persisted query via POST.  If `graphql-server` fails to find the persisted query, the runtime will fallback to sending the full query


tlgimenes:
I think it's finished by now \o/ @brunoabreu and @firstdoit, please take a look. 

Before releasing a new version based on this branch, remember to check if [this](https://github.com/vtex/graphql-server/pull/30) and [this](https://github.com/vtex/builder-hub/pull/67) were released.

react/package.json
```diff
@@ -1,16 +1,17 @@
...
+    "apollo-link-persisted-queries": "^0.1.0",
     "apollo-cache-inmemory": "^1.1.5",
     "apollo-client": "^2.2.0",
     "apollo-link-http": "^1.3.2",
     "apollo-utilities": "^1.0.4",
     "classnames": "^2.2.5",
     "eventemitter3": "^3.0.0",
     "exenv": "^1.2.2",
-    "graphql": "^0.12.3",
+    "graphql": "^0.11.0",
```
mairatma: Why use an older version of `graphql`? It's important for us to be on the most recent version as possible.

tlgimenes: Because graphql 0.12.3 conflicted with other packages. Apparently, version 0.13.1 fix this issue

tlgimenes:
Squashed all commits with a better summary, thanks @mairatma !

tlgimenes:
Upgrading to runtime 7.x because persisted queries might break on `graphql-server 0.x`. Please take a look @brunoabreu if the changes made in here reflect correctly the major upgrade

PR #29, opened 2018-03-20
Changes graphql endpoint according to context directive

tlgimenes:
Sends graphql query to endpoint /_v/v{version}/${scope}/graphql with defaults of version = 1 and scope = public

### DIFF
--- a/react/package.json
+++ b/react/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "apollo-cache-inmemory": "^1.1.5",
     "apollo-client": "^2.2.0",
+    "apollo-link-batch-http": "^1.2.1",
     "apollo-link-http": "^1.3.2",
     "apollo-link-persisted-queries": "^0.2.0",
     "apollo-utilities": "^1.0.4",
@@ -30,6 +31,7 @@
     "route-parser": "^0.0.5"
   },
   "devDependencies": {
+    "@types/graphql": "^0.12.5",
     "@types/history": "^4.6.2",
     "@types/intl": "^1.2.0",
     "@types/node": "^9.4.7",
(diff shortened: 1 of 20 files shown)


react/package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "apollo-cache-inmemory": "^1.1.5",
     "apollo-client": "^2.2.0",
+    "apollo-link-batch-http": "^1.2.1",
```
firstdoit: Is this necessary?

react/utils/client/index.ts
```diff
@@ -24,29 +25,28 @@ export const getState = (runtime: RenderRuntime) => {
 }
 
 export const getClient = (runtime: RenderRuntime) => {
-  const {graphQlUri, account, workspace} = runtime
+  const {account, workspace} = runtime
```
firstdoit: We will need the `graphQlUri` in other parts of the runtime, like fetching messages. @brunoabreu @tlgimenes how do we go around that?

react/package.json
tlgimenes: Sorry, interferences from other features, changed now

react/utils/client/index.ts
tlgimenes: I have no idea. I think you and @brunoabreu  can make a patch to this branch solving this problem

react/utils/client/generateHash.ts
```diff
@@ -0,0 +1,19 @@
...
+  Argument(node: ArgumentNode) {
+    if (node.name.value === 'hash') {
+      asset.hash = node.value.value
+    }
+  }
+})
+
+export const generateHash = (query: DocumentNode) => {
+  if (query.documentId) {
+    return query.documentId
```
brunoabreu: What is this `documentId`? Is it the same as `hash`?

tlgimenes: Yes, it should be there for legacy compatibility

PR #46, opened 2018-05-10
[WIP] updateRuntime update apps-etag

marcelotokarnia:


react/utils/runtime.graphql
```diff
@@ -1,5 +1,9 @@
 query Runtime ($page: String!, $production: Boolean!, $locale: String!, $renderVersion: String!) {
   page(page: $page, production: $production, locale: $locale, renderVersion: $renderVersion) {
     messagesJSON
+    pagesJSON
```
firstdoit: How was this working before?!

brunoabreu: 😱

marcelotokarnia: I was also shocked! 
Guess no one was using updateRuntime on anything critical.

PR #53, opened 2018-05-22
Ommits variable __typename in apollo link

tlgimenes:
This is a problem reported in https://github.com/apollographql/apollo-client/issues/1564 
with best solution found in https://github.com/apollographql/apollo-client/issues/1564#issuecomment-357492659


brunoabreu:
Please, rebase before merging

PR #56, opened 2018-05-29
Log client-side errors to Kibana

luizerasmo:
Log client-side errors to Kibana.

- Link this `render-runtime` branch.
- Link an app containing an error within an Extension Point (such that, when rendering a page with this component, it will display the message `Error rendering extension point [path]`).
- Search for the log at Kibana within `metric-renderruntime-*` or `metric-renderruntime-jserror-*`. It should display the corresponding `account`, `workspace`, `message`, `path`, `stack` and `componentStack`.

#### Types of changes

### DIFF
--- a/react/utils/logger.ts
+++ b/react/utils/logger.ts
@@ -1,0 +1,17 @@
+const KIBANA_NAMESPACE = 'renderruntime'
+const KIBANA_VALUE = 1
+
+const logEvent = (event: LogEvent) => {
+  // defaultData contains remaining fields required by sendMetric
+  const defaultData = {
+    namespace: KIBANA_NAMESPACE,
+    value: KIBANA_VALUE
+  }
+  const formattedEvent = Object.assign(defaultData, event)
+
+  if (vtex.NavigationCapture && vtex.NavigationCapture.sendMetric) {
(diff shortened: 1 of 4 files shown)


react/utils/logger.ts
```diff
@@ -0,0 +1,17 @@
+import axios from 'axios'
+import SplunkEvents from 'splunk-events'
+
+const SPLUNK_ENDPOINT = 'https://splunk-heavyforwarder-public.vtex.com:8088'
+const SPLUNK_TOKEN = '762f39fa-5679-4997-936c-add830df3231'
```
tlgimenes: Will we allow this splunk token here ? This is an open app, anyone with this token can do anything with our splunk, right ?

react/package.json
```diff
@@ -20,6 +20,7 @@
     "myvtex-sse": "^1.2.0",
     "prop-types": "^15.6.0",
     "qs": "^6.5.1",
+    "ramda": "^0.25.0",
```
firstdoit: We shouldn't add `ramda` to the runtime since it's pretty heavy. Or at least that's what we thought up until now. @brunoabreu what do you think?

PR #58, opened 2018-05-29
Migrate to new React Context API

brunoabreu:


react/utils/treePath.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export const withTreePath = <TOriginalProps extends {id: string}>(Component: ComponentType<TOriginalProps & TreePathProps>, provider: boolean = false): ComponentType<TOriginalProps> => {
+  class TreePath extends React.Component<TOriginalProps> {
+    public static contextTypes = {
+      treePath: PropTypes.string
+    }
+
+    public static childContextTypes = {
+      treePath: PropTypes.string
```
klzns: Nao entendi. Ta usando o context novo ou velho?

react/components/RenderProvider.tsx
```diff
@@ -340,15 +338,17 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     const root = page.split('/')[0]
     const editorProvider = extensions[`${root}/__provider`]
     const maybeEditable = !production && editorProvider
-      ? <ExtensionPointComponent component={editorProvider.component} props={{extensions, pages, page}}>{component}</ExtensionPointComponent>
+      ? <ExtensionPointComponent component={editorProvider.component} props={{extensions, pages, page}} treePath="">{component}</ExtensionPointComponent>
```
firstdoit: We don't need an `ExtensionPointComponent` here anymore because it doesn't HMR anymore. This can be replaced by a "withHMR" directly.

react/package.json
```diff
@@ -20,14 +20,13 @@
     "myvtex-sse": "^1.2.0",
     "prop-types": "^15.6.0",
     "qs": "^6.5.1",
-    "react": "^16.2.0",
+    "react": "^16.3.2",
```
firstdoit: Should we go to 16.4?

react/utils/treePath.tsx
brunoabreu: Dentro desse projeto somente o novo. Mas existem outras apps que pegam o treePath do contexto, como o pages-editor. Então resolvi manter a compatibilidade, da mesma forma que foi mantida para o contexto do render, no `RenderProvider`.

react/ExtensionContainer.tsx
```diff
@@ -1,26 +1,23 @@
...
-      <Fragment>
-        {children.map(id =>
-          <ExtensionPoint {...this.props} key={id} id={id} />
-        )}
-      </Fragment>
+      <RenderContext.Consumer>
+        {runtime =>
+          <TreePathContext.Consumer>
+            {({treePath}) =>
+              getDirectChildren(runtime!.extensions, treePath)
```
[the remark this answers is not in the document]
brunoabreu: Actually, that will never happen. `React.createContext` requires an initial value but many of our context fields are only available when creating an instance of `RenderProvider`. So, it was a choice between a context that is always there (`{}`), but you need to check every field, and a context that can be undefined, but in practice it isn't, because `RenderProvider` is guaranteed be there and set its value.

I thought the latter was better:
```ts
export const RenderContext = React.createContext<RenderContext | undefined>(undefined)
```

But, I just figured out I can fool Typescript this way:
```ts
export const RenderContext = React.createContext<RenderContext>({} as any)
```

I'll do that and remove these `!.`

react/ExtensionPoint.tsx
```diff
@@ -18,98 +19,37 @@ interface State {
...
+    const extension = runtime!.extensions[treePath]
+    const component = extension ? extension.component : null
+    const extensionProps = extension ? extension.props : null
 
     const props = {
       ...parentProps,
       ...extensionProps,
       params,
-      query,
+      query
```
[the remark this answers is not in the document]
brunoabreu: Sorry

react/components/RenderProvider.tsx
brunoabreu: The component is already wrapped by `withHMR`, because it is registered into `__RENDER_7_COMPONENTS__` this way. What `ExtensionPointComponent` is doing here is just grabbing the implementation from this register and rendering it.

react/package.json
brunoabreu: We can.

react/components/ExtensionPointComponent.tsx
```diff
@@ -17,32 +19,25 @@ interface State {
...
 
-export default class ExtensionPointComponent extends PureComponent<Props, State> {
-  public static contextTypes = {
-    emitter: PropTypes.object,
-    extensions: PropTypes.object,
-    fetchComponent: PropTypes.func,
-    treePath: PropTypes.string,
-  }
-
+class ExtensionPointComponent extends PureComponent<Props & RenderContextProps, State> {
```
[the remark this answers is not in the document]
brunoabreu: We don't have to.

```
export default withContext<Props>(ExtensionPointComponent)
```

But, I think it makes sense to receive it directly.

I will change that.

react/components/Link.tsx
```diff
@@ -51,13 +47,22 @@ export default class Link extends Component<Props> {
...
+    return this.props.to || this.props.page && pathFromPageName(this.props.page, pages, this.props.params) || '#'
+  }
+
   public render() {
-    const href = this.props.to || this.props.page && pathFromPageName(this.props.page, this.context.pages, this.props.params) || '#'
-    return <a href={href} {...this.props} onClick={this.handleClick} />
+    return (
+      <RenderContext.Consumer>
+        {runtime =>
+          <a href={this.getHref(runtime!.pages)} {...this.props} onClick={this.handleClick(runtime!.navigate)} />
```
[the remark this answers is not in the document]
brunoabreu: OK. The alternative is to use the `withContext` HOC.

I'll do that.

PR #62, opened 2018-06-07
Automatic cache control split

tlgimenes:
Users `cacheHints` exported by `runtime` to correctly switch the endpoint for each query

### DIFF
--- a/react/utils/client/links/uriSwitchLink.ts
+++ b/react/utils/client/links/uriSwitchLink.ts
@@ -20,12 +20,12 @@ interface OperationContext {
 
 const hashFromExtensions = ext => ext && ext.persistedQuery && ext.persistedQuery.sha256Hash
 
-export const createUriSwitchLink = (baseURI: string) =>
+export const createUriSwitchLink = (baseURI: string, workspace: string) =>
   new ApolloLink((operation: Operation, forward?: NextLink) => {
     const hash = hashFromExtensions(operation.extensions)
     const {operationType} = assetsFromQuery(operation.query)
     const protocol = canUseDOM ? 'https:' : 'http:'
-    operation.setContext(({ fetchOptions = {}, runtime: {appsEtag, cacheHints, workspace} } : OperationContext) => {
+    operation.setContext(({ fetchOptions = {}, runtime: {appsEtag, cacheHints} } : OperationContext) => {
       const {maxAge = 'LONG', scope = 'PUBLIC', version = 1} = cacheHints[hash] || {}
       const method = (scope.toLowerCase() === 'public' && operationType.toLowerCase() === 'query') ? 'GET' : 'POST'
       return {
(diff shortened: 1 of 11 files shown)


react/utils/client/links/uriSwitchLink.ts
```diff
@@ -36,17 +18,21 @@ interface OperationContext {
...
 }
 
-export const createUriSwitchLink = (baseURI: string, workspace: string) =>
+const hashFromExtensions = ext => ext && ext.persistedQuery && ext.persistedQuery.sha256Hash
+
+export const createUriSwitchLink = (baseURI: string, runtime: RenderRuntime) =>
   new ApolloLink((operation: Operation, forward?: NextLink) => {
-    const {query} = operation
-    const assets = assetsFromQuery(operation.query)
+    const {workspace, cacheHints} = runtime
```
brunoabreu: We should get `cacheHints` from context, just like `appsEtag` below. It's safer because if the runtime data gets updated with `fetchRuntime`, we will have the correct `cacheHints` here.

You must add `cacheHints` to `createRuntimeContextLink` in `RenderProvider` for that.

react/utils/client/links/versionSplitterLink.ts
```diff
@@ -1,7 +1,86 @@
...
+      }
+    }
+  },
+}) as DocumentNode
+
+const assertSingleOperation = (query: DocumentNode) => {
+  const ops = query.definitions.filter((def: OperationDefinitionNode) => operationWhiteList.find(op => op === def.operation))
+
+  if (ops.length > 1) {
+    throw new Error('Only one operation definition is allowed per query. Please split your queries in two different files')
```
brunoabreu: Is it a breaking change?

tlgimenes: Yes, but no one is using such a thing today. This applies to the case: 

```
query {
   rootField
}
query {
   rootField
}
```

react/utils/client/links/uriSwitchLink.ts
```diff
@@ -20,14 +20,13 @@ interface OperationContext {
...
-export const createUriSwitchLink = (baseURI: string, runtime: RenderRuntime) =>
+export const createUriSwitchLink = (baseURI: string) =>
   new ApolloLink((operation: Operation, forward?: NextLink) => {
-    const {workspace, cacheHints} = runtime
     const hash = hashFromExtensions(operation.extensions)
-    const {maxAge = 'LONG', scope = 'PUBLIC', version = 1} = cacheHints[hash] || {}
     const {operationType} = assetsFromQuery(operation.query)
     const protocol = canUseDOM ? 'https:' : 'http:'
-    operation.setContext(({ fetchOptions = {}, runtime: {appsEtag} } : OperationContext) => {
+    operation.setContext(({ fetchOptions = {}, runtime: {appsEtag, cacheHints, workspace} } : OperationContext) => {
```
brunoabreu: There is no `workspace` in `context`. It just contains: `appsEtag`, `cacheHints`, `components`, `extensions`, `messages` and `pages`.

PR #69, opened 2018-06-20
 Bring back react context

brunoabreu:
Finally react-apollo is supporting SSR with react context:

https://github.com/apollographql/react-apollo/blob/master/Changelog.md#216-june-19-2018


firstdoit:
(Trust your peers)

PR #77, opened 2018-06-27
Allow extension container to add a segment to treepath

firstdoit:


react/ExtensionContainer.tsx
```diff
@@ -5,15 +5,35 @@ import {RenderContext} from './components/RenderContext'
 import ExtensionPoint from './ExtensionPoint'
 import {getDirectChildren, TreePathContext} from './utils/treePath'
 
-class ExtensionContainer extends Component {
+const join = (p: string | null, c: string | null): string => {
+  if (!p) {
```
brunoabreu: What about `return [p, c].filter(id => !!id).join('/')`?

firstdoit: :exploding_head:

nice.

PR #78, opened 2018-06-28
Add auth treatment to RenderProvider

claudivanfilho:
- Add auth treament to RenderProvider

#### Screenshots or example usage
[workspace link](https://claudivan2--storecomponents.myvtex.com/account)

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [x] New feature (a non-breaking change which adds functionality)

### DIFF
--- a/react/components/MaybeAuth.tsx
+++ b/react/components/MaybeAuth.tsx
@@ -63,7 +63,12 @@ export default class MaybeAuth extends PureComponent<Props, State> {
   public render() {
     if (this.isAuthenticatedPage()) {
       const { logged, loading } = this.state
-      return <div className="flex justify-center ma4"><Loading /></div> 
+      if (loading) {
+        return <div className="flex justify-center ma4"><Loading /></div>
+      } else if (logged) {
+        return this.props.children
+      }
+      return null
     }
     return this.props.children
   }


react/components/AuthWrapper.tsx
```diff
@@ -0,0 +1,55 @@
...
+      }
+    }
+  }
+
+  public render() {
+    return this.props.children
+  }
+}
+
+export default graphql(gql`
```
brunojdo: You must do a fetch, not a graphql query

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Auth treatment and redirectioning of the store
```
andreldsa: Redirect*

react/components/NestedExtensionPoints.tsx
```diff
@@ -27,21 +28,33 @@ export default class NestedExtensionPoints extends PureComponent<Props> {
...
-      return reverse.reduce((acc: JSX.Element | null, value: string, index: number) => (
-        <ExtensionPoint
-          id={value}
-          query={query}
-          params={this.getPageParams(runtime, segments.slice(0, segments.length - index).join('/'))}>
-          {acc}
-        </ExtensionPoint>
-      ), null as JSX.Element | null)
+      return reverse.reduce((acc: JSX.Element | null, value: string, index: number) => {
+        if (breakPoint && breakPoint.point === value && breakPoint.isLoading) {
```
brunoabreu: I see there is a desirable behavior that in case of nested pages we should be able to render every "page" until we got one that requires authentication. Like rendering `store` header and footer components but not `store/account` component.

Then I think a better approach would be to have the `<AuthWrapper>` here, and not in `RenderProvider`

```js
return reverse.reduce((acc: JSX.Element | null, value: string, index: number) => (
  <AuthWrapper pages={runtime.pages} page={page} navigate={runtime.navigate}>
    <ExtensionPoint
      id={value}
      query={query}
      params={this.getPageParams(runtime, segments.slice(0, segments.length - index).join('/'))}>
      {acc}
    </ExtensionPoint>
  </AuthWrapper>
), null as JSX.Element | null)
```

Also we could rename `AuthWrapper` to `MaybeAuth`.

react/components/MaybeAuth.tsx
```diff
@@ -0,0 +1,72 @@
+import PropTypes from 'prop-types'
+import React, {Component, PureComponent} from 'react'
+import * as hotGlobals from '../core/main'
```
brunoabreu: We should have a component named `Loading`, in its own file.

react/package.json
```diff
@@ -6,6 +6,7 @@
     "lint": "tsc --noEmit && tslint src/**/*.ts"
   },
   "dependencies": {
+    "@vtex/styleguide": "5.2.3",
```
brunoabreu: Let's use the loader from our `BuildStatus`. Render runtime should not depend on styleguide, specially from `npm`. It's possible to depend on styleguide as an app (in manifest.json) and get updates automatically, but unfortunately this would create a cyclic dependency.

Later we will have a mechanism to declare a `store/__loading` extension point and use it here, then `dreamstore-theme` could use a styleguide component as the default store loading.

PR #80, opened 2018-06-29
Refactor error handling

brunoabreu:


brunoabreu:
This whole solution is actually wrong because it does not solve the case for imported components from other apps

PR #86, opened 2018-07-09
Add visibility change event listener to reconnect SSE

luizerasmo:
- Usually,  connection to SSE server fails when browser is idle for a while, and thereafter page does not automatically reload when app code changes.
- This PR adds a 'visibility change' event listener, and reconnects to SSE server every time user focus on browser screen. Thus, page automatically reloads even after being idle for a long period.

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -9,8 +9,6 @@ import {IntlProvider} from 'react-intl'
 import {History, Location, LocationListener, UnregisterCallback} from 'history'
 import {fetchAssets, getImplementation} from '../utils/assets'
 import {getClient} from '../utils/client'
-import {registerEmitter} from '../utils/events'
-import {getBaseURI} from '../utils/host'
 import {loadLocaleData} from '../utils/locales'
 import {createLocaleCookie, fetchMessages, fetchMessagesForApp} from '../utils/messages'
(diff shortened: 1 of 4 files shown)


react/components/RenderProvider.tsx
```diff
@@ -120,6 +122,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     if (!production) {
       emitter.addListener('localesUpdated', this.onLocalesUpdated)
       emitter.addListener('extensionsUpdated', this.updateRuntime)
+      document.addEventListener('visibilitychange', this.onVisibilityChanged)
```
firstdoit: I would prefer if we don't change the `RenderProvider` and instead do something like this inside the `if` that checks `emittersByWorkspace[`${account}/${workspace}`]`

```  // Share SSE connections for same account and workspace
  if (!emittersByWorkspace[`${account}/${workspace}`]) {
    emittersByWorkspace[`${account}/${workspace}`] = []
    emittersByWorkspace[`${account}/${workspace}`].eventSource = initSSE(account, workspace, baseURI)
    if (!production ) {
      document.addEventListener('visibilitychange', () => {
        if (!document.hidden && emittersByWorkspace[`${account}/${workspace}`].eventSource.readystate == ruim) {
          emittersByWorkspace[`${account}/${workspace}`].eventSource = initSSE(account, workspace, baseURI)
        }
      })
    }
  }```

PR #115, opened 2018-08-16
Add treatment of session authentication to `MaybeAuth`

ClaudivanFilho:
- Add treatment of session authentication to `MaybeAuth`.

https://claudivan--storecomponents.myvtex.com/account/orders

#### Types of changes

### DIFF
--- a/react/components/MaybeAuth.tsx
+++ b/react/components/MaybeAuth.tsx
@@ -32,10 +32,14 @@ export default class MaybeAuth extends PureComponent<Props, State> {
   }
 
   public redirectToLogin() {
-    this.props.navigate({
-      fallbackToWindowLocation: false,
-      to: LOGIN_PATH,
-    })
+    if (this.props.page !== 'store/login') {
+      const pathName = window.location.pathname.replace(/\/$/, '')
+      this.props.navigate({
+        fallbackToWindowLocation: false,
+        to: LOGIN_PATH,
+        query: `returnUrl=${pathName}`
+      })
+    }
   }
 
   public render() {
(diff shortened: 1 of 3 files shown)


react/components/MaybeAuth.tsx
```diff
@@ -57,25 +58,32 @@ export default class MaybeAuth extends PureComponent<Props, State> {
...
+        .then(response => response.json())
+        .catch(() => null)
+      const promiseAuth = fetch(AUTH_STORE_URL, { credentials: 'same-origin' })
+        .then(response => response.json())
+        .catch(() => null)
+      const [sessionRes, authRes] = await Promise.all([
+        promiseSession, promiseAuth
+      ])
+      if (
+        (sessionRes && sessionRes.namespaces.profile.isAuthenticated.value !== 'False') ||
```
brunojdo: Create a func `convertToBool`, and return `toLower(sessionRes.namespaces.profile.isAuthenticated.value) === 'true' `

react/components/MaybeAuth.tsx
```diff
@@ -2,7 +2,7 @@ import React, {PureComponent} from 'react'
 import Loading from './Loading'
 
 const LOGIN_PATH = '/login'
-const AUTH_STORE_URL = '/_v/private/authenticated/store'
+const API_SESSION_URL = '/api/sessions?items=*'
```
brunojdo: You need to make a `POST /api/sessions` with a header `Content-Type: application/json` to initialize session . (retrieve `vtex_session` and `vtex_segment` cookies)

brunojdo: PS: Check if you already have these cookies, if not make the request.

brunoabreu: Actually `vtex_session` is (or should be) an HTTP only cookie, so there is no way to know if there cookie is there using javascript.

PR #116, opened 2018-08-16
Handling rendering inside iframe

pedroteosousa:


react/components/RenderProvider.tsx
```diff
@@ -80,6 +84,15 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     runtime: PropTypes.object,
   }
 
+  public sendInfoFromIframe = debounce(() => {
+    const context = this.getChildContext()
+    if (window.top !== window.self) {
```
firstdoit: For the future: we should have an easier `isIframe` function available

firstdoit: Oh, and we should also check for the existence of `window.top.__provideRuntime`

react/components/RenderProvider.tsx
```diff
@@ -476,22 +498,18 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
       <RenderContext.Provider value={context}>
         <TreePathContext.Provider value={{treePath: ''}}>
           <ApolloProvider client={this.apolloClient}>
             <IntlProvider locale={locale} messages={mergedMessages}>
               <Fragment>
-                {!production && <BuildStatus />}
-                {maybeEditable}
+                {!production && !isIframe && <BuildStatus />}
+                {component}
+                { isIframe ? <ExtensionPoint id="store/__overlay" /> : null}
```
firstdoit: For the future: this should move to `vtex.store`

react/components/RenderProvider.tsx
firstdoit: And we could probably only incur in the `getChildContext` call after we're sure we're sending it.

PR #125, opened 2018-08-27
Should use post on persisted query not found

tlgimenes:
Currently, when a `PersistedQueryNotFound` error is thrown, the system redos the query using the same HTTP verb with the whole stringified version of the query. If it turned out to be a public graphql query, the whole query text will be stringified into the request's querystring, putting too much pressure into the browser and our systems. 

This PR solves this issue by using GET in the first fetch, and POST in case of error

### DIFF
--- a/react/utils/client/links/uriSwitchLink.ts
+++ b/react/utils/client/links/uriSwitchLink.ts
@@ -49,9 +49,9 @@ const extractHints = (query: ASTNode, meta: CacheHints) => {
 
 export const createUriSwitchLink = (baseURI: string, workspace: string) =>
   new ApolloLink((operation: Operation, forward?: NextLink) => {
-    operation.setContext(({ fetchOptions = {}, runtime: {appsEtag, cacheHints} } : OperationContext) => {
-      const oldContext = operation.getContext()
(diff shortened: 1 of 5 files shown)


react/utils/client/links/uriSwitchLink.ts
```diff
@@ -50,11 +50,14 @@ const extractHints = (query: ASTNode, meta) => {
 export const createUriSwitchLink = (baseURI: string, workspace: string) =>
   new ApolloLink((operation: Operation, forward?: NextLink) => {
     operation.setContext(({ fetchOptions = {}, runtime: {appsEtag, cacheHints} } : OperationContext) => {
+      const oldContext = operation.getContext()
```
brunoabreu: **nit**: the old context is already available as `setContext` param. I think something like this would be clearer:

```
operation.setContext((oldContext: OperationContext) => {
  const {fetchOptions = {}, runtime: {appsEtag, cacheHints}} = oldContext
  const oldMethod = fetchOptions.method || 'POST'
  ...
  ...
})
```

tlgimenes: thanks @brunoabreu !!

PR #127, opened 2018-08-29
Clear apps errors when navigate

estacioneto:
Only HMR events were cleaning errors, but the user needs to be allowed to navigate through the store without having to reload the page. So, when the ExtensionPoints receives new props, the errors are cleaned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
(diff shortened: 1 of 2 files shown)


estacioneto:
Actually, @firstdoit, I think it already gets the different errors. The cycle is:
1. Error occurs;
2. `componentDidCatch` is called;
3. Error Component is mounted;
4. `componentDidUpdate` is called and sets `this.mountedError = true`;
5. Something changes and the component receives new props;
6. Error Component rendered again;
7. `componentDidUpdate` is called and clears the error;
8. Tries to mount the app again (if there's any error, different or not, the `componentDidCatch` will be called again).

And, about `prettier`, I tried to do this putting the format change only in the first commit but I get this is annoying anyway :(

PR #129, opened 2018-09-05
Adds batched queries support

tlgimenes:
This PR not only adds batched queries support, but also defines a standard for sending batched queries in a `GET` request.

The usual way of sending a GraphQL query in an http request is by parsing the query object into the url's querystring. For instance, suppose we have the following query object:

```
{
'operationName': 'GetProducts',
'query': 'query GetProducst {products {slug}}',
'variables': {}
}
```

The corresponding url's querystring for this request is:

`?operationName='GetProducts'&query='query GetProducst {products {slug}}'&variables={}'`

The problem in using this approach for sending batched queries is that the query object is an array, and the url's querystring does not support array objects. To overcome this problem the batched query object should not be parsed directly, but rather inserted under a key named `batch`. This key can be repeated many times, each one indicating a new query object. Also, to avoid duplication of keys, the query object fields are separated by `@@` in the following order: operationName, variables and extensions. To illustrate how the stringification works, suppose we have following query objects

```
[{
'operationName': 'GetProducts1',
'extensions': 'extensions: { persistedQuery{ sha256: "3123123123123"}}',
'variables': {}
},
{
'operationName': 'GetProducts2',
'extensions': 'extensions{ persistedQuery{ sha256: "gfdgdfgdfgdfg"}}',
'variables': {}
}]
```

The corresponding url's querystring for this request is:
`?batch='GetProducts1'@@{}@@extensions{ persistedQuery{ sha256: "gfdgdfgdfgdfg"}}'&batch=batch='GetProducts1'@@{}@@extensions{ persistedQuery{ sha256: "3123123123123"}}`

This way, the query is correctly sent and `graphql-server` can parse back the query object

### DIFF
--- a/react/package.json
+++ b/react/package.json
@@ -29,8 +29,7 @@
     "react-hot-loader": "^4.1.2",
     "react-intl": "^2.4.0",
     "react-no-ssr": "^1.1.0",
-    "route-parser": "^0.0.5",
-    "url": "^0.11.0"
+    "route-parser": "^0.0.5"
   },
   "devDependencies": {
     "@types/debounce": "^1.0.0",
--- a/react/package.json
+++ b/react/package.json
@@ -8,7 +8,7 @@
   "dependencies": {
     "apollo-cache-inmemory": "^1.2.5",
     "apollo-client": "^2.3.5",
-    "apollo-link-http": "^1.5.4",
+    "apollo-link-batch-http": "^1.2.2",
     "apollo-link-persisted-queries": "^0.2.1",
     "apollo-upload-client": "^8.1.0",
     "apollo-utilities": "^1.0.16",
@@ -29,7 +29,8 @@
     "react-hot-loader": "^4.1.2",
     "react-intl": "^2.4.0",
     "react-no-ssr": "^1.1.0",
-    "route-parser": "^0.0.5"
+    "route-parser": "^0.0.5",
+    "url": "^0.11.0"
   },
   "devDependencies": {
     "@types/debounce": "^1.0.0",
--- a/react/package.json
+++ b/react/package.json
@@ -29,7 +29,9 @@
     "react-hot-loader": "^4.1.2",
     "react-intl": "^2.4.0",
     "react-no-ssr": "^1.1.0",
-    "route-parser": "^0.0.5"
+    "route-parser": "^0.0.5",
+    "url": "^0.11.0",
+    "url-search-params": "^1.1.0"
   },
   "devDependencies": {
     "@types/apollo-upload-client": "^8.1.0",
@@ -46,6 +48,7 @@
     "@types/react-hot-loader": "^4.1.0",
     "@types/react-intl": "^2.3.5",
     "@types/route-parser": "^0.1.1",
+    "@types/url-search-params": "^0.10.2",
     "tslint": "^5.2.0",
     "tslint-config-vtex": "^2.0.0",
     "tslint-eslint-rules": "^5.1.0",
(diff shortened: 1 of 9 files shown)


react/package.json
```diff
@@ -29,7 +29,8 @@
     "react-hot-loader": "^4.1.2",
     "react-intl": "^2.4.0",
     "react-no-ssr": "^1.1.0",
-    "route-parser": "^0.0.5"
+    "route-parser": "^0.0.5",
+    "url": "^0.11.0"
```
firstdoit: No need to add this package

react/utils/client/batchKey.ts
```diff
@@ -0,0 +1,22 @@
+import { Operation } from 'apollo-link'
+import { BREAK, DocumentNode, FieldNode, visit } from 'graphql'
+
+const assetsFromQuery = (query: DocumentNode) => {
+  const assets = {name: ''}
+  visit(query, {
+    Field(node: FieldNode, _, __, path) {
+      if (path.length === 5) {
```
firstdoit: Make this into a const and comment, please :)

react/utils/client/fetch.ts
```diff
@@ -0,0 +1,21 @@
...
+export const IOFetch = async (uri: string | Request, init?: RequestInit): Promise<Response> => {
+  const {useGetInBatch = false, body} = init || {} as any
+  const formattedInit = {...init}
+  let formattedUri = uri
+
+  if (useGetInBatch && typeof uri === 'string' && formattedInit) {
+    const parsedUri = parse(uri, true)
+
+    delete parsedUri.search
+    parsedUri.query = {...parsedUri.query, batch: body}
```
firstdoit: Let's make this into a `batch[]={}&batch[]={}&...` so we use rails' and PHP format.

react/utils/client/fetch.ts
```diff
@@ -0,0 +1,21 @@
+import {format, parse} from 'url'
+
+export const IOFetch = async (uri: string | Request, init?: RequestInit): Promise<Response> => {
+  const {useGetInBatch = false, body} = init || {} as any
+  const formattedInit = {...init}
+  let formattedUri = uri
+
+  if (useGetInBatch && typeof uri === 'string' && formattedInit) {
```
firstdoit: Check for Request: https://github.com/vtex/render-server/blob/master/node/middleware/render/sandbox.ts#L24

react/utils/client/index.ts
```diff
@@ -78,9 +82,11 @@ export const getClient = (runtime: RenderRuntime, baseURI: string, runtimeContex
 
     const uriSwitchLink = createUriSwitchLink(baseURI, workspace)
 
+    const baseLink = [omitTypenameLink, versionSplitterLink, runtimeContextLink, ensureSessionLink, persistedQueryLink, uriSwitchLink]
+
     const link = cacheControl
-      ? ApolloLink.from([omitTypenameLink, versionSplitterLink, runtimeContextLink, ensureSessionLink, persistedQueryLink, uriSwitchLink, cachingLink(cacheControl), fetcherLink])
-      : ApolloLink.from([omitTypenameLink, versionSplitterLink, runtimeContextLink, ensureSessionLink, persistedQueryLink, uriSwitchLink, fetcherLink])
+      ? ApolloLink.from([...baseLink, cachingLink(cacheControl), fetcherLink])
```
firstdoit: Maybe create a separate array for `cachingLink` which optionally has this link or is empty, and always spread it here. (So we avoid two similar arrays)

PR #134, opened 2018-09-19
Infinity calls when using fetchWithRetry in initializeSession and patchSession.

salesfelipe:
Fix bug ifinite calls

[workspace](https://felipe--storecomponents.myvtex.com/admin/cms/storefront)

#### Types of changes
- [X] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)


react/utils/session.tsx
```diff
@@ -26,7 +26,7 @@ const fetchWithRetry = (url: string, init: RequestInit, maxRetries: number = 3):
         return delay(ms)
           .then(() => callFetch(++attempt))
       }
-    })
+    }).catch((error) => console.error(error))
```
brunoabreu: A better approach would be:

- throw an error on the `then` above if there is an error
- move the retry code to `catch`

PR #157, opened 2018-10-23
Above the fold and preview props on LayoutContainer

athoscouto:
Completing @firstdoit work on above the fold.

If the `aboveTheFold : number` prop is set on any `LayoutContainer` it'll render only it's first `aboveTheFold` elements on the server side. Once the component is rendered on client side it'll render the other elements.

The last commit introduces the `preview : boolean` prop. If set to `true`, it won't render elements that are below the fold on the client side.

@brunoabreu and @tlgimenes, I'm not sure we'll need the preview prop. I've added it anyway and if we don't need it I'll just strip the commit.

### DIFF
--- a/react/components/LayoutContainer.tsx
+++ b/react/components/LayoutContainer.tsx
@@ -3,27 +3,49 @@ import React, { Component } from 'react'
 
 import ExtensionPoint from './ExtensionPoint'
 
-type Element = string | any[]
+type Element = string | ElementArray
+interface ElementArray extends Array<Element> {}
 
 interface LayoutContainerProps {
-  elements: Element[]
+  aboveTheFold?: number
+  elements: Element
 }
 
 interface ContainerProps {
+  aboveTheFold?: number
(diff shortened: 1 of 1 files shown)


react/components/LayoutContainer.tsx
```diff
@@ -61,7 +66,7 @@ class Container extends Component<ContainerProps, ContainerState> {
 
     const returnValue: JSX.Element[] = elements.slice(0, this.state.elementsToRender).map((element: Element) => {
       return (
-        <Container key={element.toString()} elements={element} isRow={!isRow} {...props} aboveTheFold={undefined} >
+        <Container key={element.toString()} elements={element} isRow={!isRow} {...props} aboveTheFold={undefined} preview={undefined}>
```
firstdoit: You can simply extract `aboveTheFold` and `preview` props in the destructuring of line 53 instead of explicitly passing undefined.

PR #167, opened 2018-11-16
Use custom bluebird for SSR

tlgimenes:
This PR uses a custom bluebird instance in SSR. This is necessary for not messing around with render's promises and ultimately throwing unecessary `unhandledPromiseRejection` errors in render-server

### DIFF
--- a/public/externals.json
+++ b/public/externals.json
@@ -2,6 +2,10 @@
   "commons": {
     "dev": [
       {
+        "path": "https://unpkg.com/bluebird@3.5.1/js/browser/bluebird.js",
+        "serverOnly": true
+      },
+      {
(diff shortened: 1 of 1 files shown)


public/externals.json
```diff
@@ -1,6 +1,10 @@
 {
   "commons": {
     "dev": [
+      {
+        "path": "https://unpkg.com/bluebird@3.5.1/js/browser/bluebird.js",
```
firstdoit: Is the `browser` version the one we want for this case? I guess it overrides window.Promise, right? In this case, LGTM

tlgimenes: yes, it overrides window.Promise :)

PR #169, opened 2018-11-21
Scroll element into view when elementId is passed to scroll options

athoscouto:
Allow scroll after navigate to target elements and not just absolute positions.

### DIFF
--- a/react/typings/global.d.ts
+++ b/react/typings/global.d.ts
@@ -78,7 +78,11 @@ declare global {
     id: string
   }
 
-  type RenderScrollOptions = ScrollToOptions | false
+  interface RenderScrollToOptions extends ScrollToOptions {
+    elementId?: string
+  }
+
+  type RenderScrollOptions = RenderScrollToOptions | false
 
   interface RenderHistoryLocation extends Location {
     state?: {
--- a/react/typings/global.d.ts
+++ b/react/typings/global.d.ts
@@ -78,11 +78,11 @@ declare global {
     id: string
   }
 
-  interface RenderScrollToOptions extends ScrollToOptions {
-    elementId?: string
+  interface RelativeScrollToOptions extends ScrollToOptions {
+    baseElementId?: string
   }
 
-  type RenderScrollOptions = RenderScrollToOptions | false
+  type RenderScrollOptions = RelativeScrollToOptions | false
 
   interface RenderHistoryLocation extends Location {
     state?: {
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -1,7 +1,6 @@
 import {canUseDOM} from 'exenv'
 import {History, LocationDescriptorObject} from 'history'
 import * as RouteParser from 'route-parser'
-import { element } from 'prop-types';
 
 const EMPTY_OBJECT = Object.freeze && Object.freeze({}) || {}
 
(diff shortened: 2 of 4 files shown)


react/typings/global.d.ts
```diff
@@ -78,7 +78,11 @@ declare global {
     id: string
   }
 
-  type RenderScrollOptions = ScrollToOptions | false
+  interface RenderScrollToOptions extends ScrollToOptions {
```
victorsenam: `ScrollToOptions` is named like that because it directly reffers [the one in the Web API](https://developer.mozilla.org/en-US/docs/Web/API/ScrollToOptions). `RenderScrollOptions` is named like that because we think `RenderScrollToOptions` is ugly.

I don't think having both `RenderScrollToOptions` and `RenderScrollOptions` is a great idea because they both have the exact same meaning. I would use something like `ExtendedScrollToOptions` or `PagesScrollToOptions` and leave the rest as it is.

react/utils/pages.ts
```diff
@@ -125,7 +126,16 @@ export function navigate(history: History | null, pages: Pages, options: Navigat
   return false
 }
 
-export function scrollTo(options: ScrollToOptions) {
+export function scrollTo(options: RenderScrollToOptions) {
+  const { elementId } = options
+  if (elementId) {
+    const scrollAnchor = document.querySelector(`#${elementId}`)
+    if (scrollAnchor) {
```
victorsenam: I do not love the fact that `elementId`, when set, ignores the other options. As I've commented in person, I would use [`element.getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect), `window.scrollX` and `window.scrollY` to still allow offsets.

Alternatively, I would make `RenderScrollToOptions` not be an extension of `ScrollToOptions`, but an union, which makes it clear that either one or the other configuration may be applied:

```
type AchoredScrollOptions = { elementId: string }
type RenderScrollToOptions = ScrollToOptions | AnchoredScrollOptions
``` 

Still taking into account my name change suggestion in the comment above.

athoscouto:
Addressed comments. Feel free to test it at https://athos--boticario.myvtex.com/perfumaria/d

react/typings/global.d.ts
athoscouto: Agree

react/utils/pages.ts
athoscouto: Made the scroll relative to the element with the ID

react/utils/pages.ts
```diff
@@ -125,14 +125,29 @@ export function navigate(history: History | null, pages: Pages, options: Navigat
...
+  }
+
+  const { top, left } = scrollAnchor.getBoundingClientRect()
+  polyfillScrollTo({
+    left: left + window.scrollX + (options.left || 0),
+    top: top + window.scrollY + (options.top || 0)
+  })
+}
+
+function polyfillScrollTo(options: ScrollToOptions) {
```
victorsenam: The `polyfill` part is just the thing inside the `catch`, but no one cares, I'm just being pedantic.

athoscouto:
@victorsenam https://github.com/vtex-apps/render-runtime/pull/169/files#diff-d441393263ad8d3eb963aba81b164203R186
I think I did .-.
Anything wrong?

PR #170, opened 2018-11-26
Update vtex-tachyons@2.9.0

klzns:
I linked render-runtime with this code, but vtex-tachyons didn't change to the new version. It somehow still got the v2.8.0.

Anyway, that's what I usually do to update it.


brunojdo:
It was updated by @firstdoit

PR #180, opened 2018-12-03
Fix workspace not being found when using vteximg

arthurepc:


react/utils/assets.ts
```diff
@@ -11,9 +11,9 @@ export const getVTEXImgHost = (account: string) => {
   return `https://${account}.vteximg.com.br`
 }
 
-const getAbsoluteURL = (account: string, url: string) => {
-  return isRelative(url)
-    ? `${getVTEXImgHost(account)}${url}`
+const getAbsoluteURL = (account: string, url: string, workspace: string, production: boolean) => {
+  return isRelative(url) && production
+    ? `${getVTEXImgHost(account)}${url}` + `?workspace=${workspace}`
```
tlgimenes: why not using the already existent string template like so: 
```suggestion
    ? `${getVTEXImgHost(account)}${url}?workspace=${workspace}`
```

react/utils/assets.ts
```diff
@@ -130,8 +130,8 @@ export function getExtensionImplementation<P={}, S={}>(extensions: Extensions, n
   return extension && extension.component ? getImplementation<P, S>(extension.component) : null
 }
 
-export function fetchAssets(account: string, assets: string[]) {
-  const absoluteAssets = assets.map(url => getAbsoluteURL(account, url))
+export function fetchAssets(runtime: RenderRuntime, assets: string[]) {
+  const absoluteAssets = assets.map(url => getAbsoluteURL(runtime.account, url, runtime.workspace, runtime.production))
```
tlgimenes: isn't it clearer destructuring before using the variables ?
```suggestion
  const { account, workspace, production } = runtime
```

PR #184, opened 2018-12-10
Fix/bugs from previous versions

luizerasmo:
This PR follows #182 and #183 (which attempted to fix the infinite loop issue when `render-runtime` was unable to get the `component`'s implementation). The current PR aims at solving a few bugs / issues that were tracked during the past week's investigation:
- Update the `this.state` `components` prop properly. Previously, we could 'lose' some of the components during navigation (because the `setState` update did not account for the previous state's `components` prop). This was causing an issue in the `feed` account.
- Do not account for `workspace` when fetching assets from `vteximg` host. The asset is already determined by the app's name and version (this was causing some assets to be loaded twice, because the URL with a querystring implied a 'new' script in the page). This was observed in `gocommerce`'s admin when the workspace was set at `production` mode.

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -387,7 +387,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
       this.setState({
         appsEtag,
         cacheHints,
-        components,
+        components: {...this.state.components, ...components},
         extensions,
         messages,
         page,
@@ -431,7 +431,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     }))
 
     this.setState(({components, messages}) => ({
-      components: {...defaultComponents, ...components},
+      components: {...defaultComponents, ...this.state.components, ...components},
       defaultExtensions,
       messages: {...defaultMessages, ...messages}
     }))
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -387,9 +387,9 @@ class RenderProvider extends Component<Props, RenderProviderState> {
       this.setState({
         appsEtag,
         cacheHints,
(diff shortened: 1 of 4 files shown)


react/utils/assets.ts
```diff
@@ -12,9 +12,9 @@ export const getVTEXImgHost = (account: string) => {
   return `https://${account}.vteximg.com.br`
 }
 
-const getAbsoluteURL = (account: string, url: string, workspace: string, production: boolean) => {
+const getAbsoluteURL = (account: string, url: string, production: boolean) => {
   return isRelative(url) && production
-    ? `${getVTEXImgHost(account)}${url}?workspace=${workspace}`
+    ? `${getVTEXImgHost(account)}${url}`
```
tlgimenes: If the workspace is in production, we should fetch the corresponding asset, right ?

react/components/RenderProvider.tsx
```diff
@@ -387,7 +387,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
       this.setState({
         appsEtag,
         cacheHints,
-        components,
+        components: {...this.state.components, ...components},
```
tlgimenes: shouldn't we add extensions and messages as well ?

react/components/RenderProvider.tsx
```diff
@@ -431,7 +431,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     }))
 
     this.setState(({components, messages}) => ({
-      components: {...defaultComponents, ...components},
+      components: {...defaultComponents, ...this.state.components, ...components},
       defaultExtensions,
       messages: {...defaultMessages, ...messages}
```
tlgimenes: we should use messages here as well, right ?

PR #216, opened 2019-02-06
Add Support to React Hooks!

brunojdo:
React hooks!!! 🎉 

Also, create our first hook: `useRuntime`. Thanks @lucasecdb


react/typings/global.d.ts
```diff
@@ -341,9 +339,12 @@ interface RenderComponent<P={}, S={}> {
     withHMR: any
     withRuntimeContext: any
     withSession: any
+    useRuntime: any
```
klzns: Change it to

```js
useRuntime(): RenderRuntime
```

PR #217, opened 2019-02-06
Render SDK 8

brunojdo:
### Added
- Apollo Link State support.
- React 16.8.0, Hooks, profile... 🎉 
- Create hook `useRuntime` in favor of `withRuntimeContext`. 
- Add goBack method to the provider

### Changed
- Now, import from `render` is `vtex.render-runtime`.
- Adapt to use new blocks language to build pages.
- Loading SVG is loaded by Icon Pack. 
- Change the position of the building indicator when it has hovered.
- Get context as extension property instead of `__context`.
- Add wrappers when loading preview

### Removed
- Support for legacy extensions.
- Support for old pages protocol.
- `MaybeAuth` component.


luciannojunior:
Cool

PR #222, opened 2019-02-13
Fix direct child not working with '$'

thiagomurakami:
The dollar sign is used in some special blocks names (`$after`, `$before`, `$around`) and the Regexp for detecting a direct child wasn't working due to `$` being a regex modifier.

Also, configure `jest` on this project.

### DIFF
--- a/react/jest.config.js
+++ b/react/jest.config.js
@@ -10,7 +10,6 @@ module.exports = {
     "js",
     "json"
   ],
-  // "setupFilesAfterEnv": "<rootDir>/testUtils/setupTests.js",
   "moduleDirectories": [
     "node_modules",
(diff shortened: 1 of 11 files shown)


react/jest.config.js
```diff
@@ -0,0 +1,27 @@
...
+  "transform": {
+    "^.+\\.tsx?$": "ts-jest"
+  },
+  "moduleFileExtensions": [
+    "ts",
+    "tsx",
+    "js",
+    "json"
+  ],
+  // "setupFilesAfterEnv": "<rootDir>/testUtils/setupTests.js",
```
alcar: Commented-out code.

PR #225, opened 2019-02-14
Feature/Send events passed in track field of interfaces

FMota0:
Add invisible wrapper to send events that are passed in track field in interfaces.json

The interfaces gonna have the possibility of adding events that should be captured to be sent to request capture, add logic of sent in wrapper during render

#### Types of changes

### DIFF
--- a/react/components/TrackEventsWrapper.tsx
+++ b/react/components/TrackEventsWrapper.tsx
@@ -21,13 +21,18 @@ const sendEvent = (id: string, event: string) => {
 
 class TrackEventsWrapper extends PureComponent<Props> {
 
+  public eventsHandler: Record<string, () => void> = {}
+
   public componentDidMount(){
     const { id, events } = this.props
     if(events && events.length && id){
       const element = ReactDOM.findDOMNode(this)
       if(element && element.addEventListener){
         forEach(
-          event => element.addEventListener(event, () => sendEvent(id, event))
+          event => {
+            this.eventsHandler[event] = () => sendEvent(id, event)
+            element.addEventListener(event, this.eventsHandler[event])
+          }
         , events)
       }
     }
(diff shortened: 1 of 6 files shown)


react/components/TrackEventsWrapper.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  public componentWillUnmount(){
+    const { id, events } = this.props
+    if(events && events.length && id){
+      const element = ReactDOM.findDOMNode(this)
+      if(element && element.addEventListener){
+        forEach(
+          event => element.removeEventListener(event, () => sendEvent(id, event))
```
lucasecdb: this won't work, because you are creating a new function every step of the loop, and these won't be the same reference as the one you registered with `addEventListener`.

instead you should create a property on this class with the event handler and register and unregister passing in the same function. you could use [this class](https://github.com/vtex/pixel-manager/blob/951350bd1365c52903c9d4aea0463628a344d331/react/PixelIFrame.tsx#L21-L49) as a reference

react/components/TrackEventsWrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+class TrackEventsWrapper extends PureComponent<Props> {
+
+  public eventsHandler: Record<string, () => void> = {}
+
+  public componentDidMount(){
+    const { id, events } = this.props
+    if(events && events.length && id){
+      const element = ReactDOM.findDOMNode(this)
```
lucasecdb: can we find another way around this without the use of `findDOMNode`? This method is getting deprecated and have several issues, specially regarding component update.

FMota0: I searched, but didn't find anything that could replace it in this case

FMota0: It's already being used in https://github.com/vtex-apps/render-runtime/blob/master/react/components/ExtensionPoint.tsx#L176

PR #229, opened 2019-02-18
Add renderExtension function to __RENDER_8_RUNTIME__

tiagonapoli:
- `renderExtension`: Given an `HTMLElement` and an `ExtensionName` the extension will be rendered using React Portals on the given `HTMLElement`. This feature will replace `RenderExtensionLoader`.

### DIFF
--- a/react/components/ExtensionManager.tsx
+++ b/react/components/ExtensionManager.tsx
@@ -1,0 +1,63 @@
+import PropTypes from 'prop-types'
+import * as R from 'ramda'
+import React, { Component } from 'react'
+
+import ExtensionPortal from './ExtensionPortal'
+
+interface Props {
+  runtime: RenderRuntime,
+}
+
+interface State {
+  extensions: SimplifiedExtension[]
+}
+
+const addOrUpdate = (extensions: SimplifiedExtension[], extension: SimplifiedExtension): SimplifiedExtension[] => {
+  const exists = R.reduce((acc, el: SimplifiedExtension) => {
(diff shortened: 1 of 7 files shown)


react/components/ExtensionManager.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface Props {
+  runtime: RenderRuntime,
+}
+
+interface State {
+  extensionsToRender: PortalRenderingRequest[]
+}
+
+const addOrUpdate = (extensionsList: PortalRenderingRequest[], newExtension: PortalRenderingRequest): PortalRenderingRequest[] => {
```
firstdoit: Could be a private method 🤷‍♀️

react/components/ExtensionPortal.tsx
```diff
@@ -0,0 +1,29 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+import {createPortal as reactCreatePortal} from 'react-dom'
```
firstdoit: Why the rename?

react/components/ExtensionManager.tsx
tiagonapoli: Yeah, you are right, done on 2711f32

react/components/ExtensionPortal.tsx
tiagonapoli: I noticed now that it was indeed unnecessary, removed on 9263185

PR #237, opened 2019-02-25
Remove unnecessary object on setState

tiagonapoli:
According to [React documentation](https://reactjs.org/docs/react-component.html#setstate), when `setState` receives a function, the first argument on that function call will be the object state in the moment that updater function was called.

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -471,9 +471,9 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     }))
 
     this.setState(({components, messages}) => ({
-      components: {...defaultComponents, ...this.state.components, ...components},
+      components: {...defaultComponents, ...components},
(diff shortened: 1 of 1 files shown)


react/components/RenderProvider.tsx
```diff
@@ -471,9 +471,9 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     }))
```
salesfelipe: Changelog?

react/components/RenderProvider.tsx
```diff
@@ -471,9 +471,9 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
 
     this.setState(({components, messages}) => ({
```
salesfelipe: Lint?
```suggestion
    this.setState(({ components, messages }) => ({
```

PR #258, opened 2019-03-25
Disable user land

Jeymisson:
To disable user blocks query string `disableUserLand`, is passed through navigation

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -374,7 +374,13 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     const transientRoute = {...route, ...navigationRoute}
     const {[page]: {allowConditions, declarer}} = pagesState
     const shouldSkipFetchNavigationData = !allowConditions && loadedPages.has(page)
-    const query = parse(location.search.substr(1))
+    const query = parse(window.location.search.substr(1))
+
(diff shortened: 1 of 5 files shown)


react/components/RenderProvider.tsx
```diff
@@ -374,7 +374,18 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     const transientRoute = {...route, ...navigationRoute}
     const {[page]: {allowConditions, declarer}} = pagesState
     const shouldSkipFetchNavigationData = !allowConditions && loadedPages.has(page)
-    const query = parse(location.search.substr(1))
+    const query = parse(window.location.search.substr(1))
+
+    // Store and pass disableUserLand logic to navigation
+    if(this.state.query && 'disableUserLand' in this.state.query && this.state.query.disableUserLand !== 'false'){
+      query['disableUserLand'] = this.state.query.disableUserLand
+      const queryString = '?' + Object.keys(query).map(key => `${key}${query[key]? '=' + query[key]: ''}`).join('&')
```
firstdoit: Isn't there a native way to do this? Like URL()

Jeymisson: Don't think so, but I found we use https://github.com/ljharb/qs to do it. So the solution goes like this: `stringify(query)`

PR #267, opened 2019-04-02
Rename disable userland

Jeymisson:
Rename `disableUserLand` query string to `__disableUserLand` to match `__disableSSR` pattern
It should be merged after [this](https://github.com/vtex/pages-graphql/pull/161)

### DIFF
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -3,6 +3,7 @@ import { History, LocationDescriptorObject } from 'history'
 import queryString from 'query-string'
 import * as RouteParser from 'route-parser'
 
+const DISABLE_USER_LAND_KEY = '__disableUserLand'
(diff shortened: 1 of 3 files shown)


react/utils/pages.ts
```diff
@@ -245,6 +245,15 @@ function routeIdFromPath(path: string, routes: Pages) {
   return id
 }
 
+export function warnUserLand(){
+  const runtime = window.__RUNTIME__
+  const shouldWarnUserLand = !runtime.production && runtime.query
```
luciannojunior: Do something like the `mergePersistingQueries` to avoid rewriting the name of the query

luciannojunior:
Just remember to add some warning for the other case, where there is user's custom templating and `disableUserLand` is not being used 😄

PR #272, opened 2019-04-11
Removing check for the back navigate and add scrollOptions option

luciannojunior:
Related PR that should be released firstly: [store](https://github.com/vtex-apps/store/pull/211)

Test [here](https://cafedovento--storecomponents.myvtex.com/) using our search, changing the sort option, and going back. It didn't work before.

This is part of the mecanism to make possible the user navigate to `/shoes`, _shoes_ being a category, and get that page relative to a _category_. When someone accesses `/shoes/c` for the first time we save that registry and `/shoes` is now a category path, and we redirect now a user that tries to navigate to `/shoes/c`.

But that means that we would have to have two navigations for a simple transition, and with every navigation we fetch some heavy data on `pages-graphql`. To avoid fetching the same data again unnecessarily, there was the `if (!state.navigationRoute)` in the `RenderProvider`'s `onPageChanged` that would do that. But, in some cases when going back in the search-result flow (like trying to go to the last applied filter), the lack of the `navigationRoute` caused by that _canonical replace_ would make the `onPageChanged` return and not really change anything on page.

A couple of weeks ago I've added a `fetchPage` option to avoid fetching navigation data on some occasions. What I've done here is use that option instead of not passing `navigationRoute`

**FEAT.**
For the _runtime_, I took advantage and also have added the `scrollOptions` to `setQuery`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - Support for `scrollOptions` on `setQuery`.
+
+### Changed
 - Removing `navigationRoute` check for the back button functionality.
 
 ## [8.18.1] - 2019-04-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Support for `scrollOptions` on `setQuery`.
+- Removing `navigationRoute` check for the back button functionality.
```
lucasecdb: maybe should be in a `Fixed` section?

PR #277, opened 2019-04-25
Switches to post when max querystring is exceeded

tlgimenes:
Currently, when the querystring is HUGE (>4096) characters, our http infrastructure bugs. To prevent this problem, when the payload size is HUGE we send the request as a POST

### DIFF
--- a/react/utils/client/links/base64Link.ts
+++ b/react/utils/client/links/base64Link.ts
@@ -1,8 +1,6 @@
 import { ApolloLink, NextLink, Operation } from 'apollo-link'
 import { Base64 } from 'js-base64'
 
-import { OperationContext } from './uriSwitchLink'
-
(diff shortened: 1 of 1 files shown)


react/utils/client/links/base64Link.ts
```diff
@@ -1,6 +1,10 @@
 import { ApolloLink, NextLink, Operation } from 'apollo-link'
 import { Base64 } from 'js-base64'
 
+import { OperationContext } from './uriSwitchLink'
+
+const MAX_QUERYSTRING_SIZE = 4096
```
marcoskwkm: Is there some reasoning behind the value `4096`?

tlgimenes: No, it's just an empirical value

PR #284, opened 2019-05-06
Prefix rootPath in links and navigation

firstdoit:


react/utils/pages.ts
```diff
@@ -162,6 +163,10 @@ export function navigate(
     return false
   }
 
+  if (rootPath && !navigationRoute.path.startsWith('http')) {
```
klzns: Add comment or something to indicate that you're checking if it's absolute

react/components/Link.tsx
```diff
@@ -57,13 +58,16 @@ const Link: React.FunctionComponent<Props> = ({
 
   const getHref = () => {
     if (to) {
+      if (rootPath && !to.startsWith('http') && !to.startsWith(rootPath)) {
```
klzns: Same

luciannojunior:
Update README.md with the new `rootPath` param

PR #288, opened 2019-05-08
Fix route receiving invalid params

mariana-mendes:

as title says.

When passing params not specified in [routes template](https://github.com/vtex-apps/store/blob/master/store/routes.json), none of warn is displayed. 

1. [Acess this workspace](https://mariana--storecomponents.myvtex.com)
2. Create a navigate object passing more params then necessary e.g `navigate({page: 'store.product',
  params: { slug: 'traveler-backpack', batata: 'batata', banana: 'banana', a:'a'},})`
3. And check if console throws this warn : 
`pages.ts:116 The following params are invalid:
batata
banana
a`

#### Types of changes

* [X] Bug fix (a non-breaking change which fixes an issue)
* [ ] New feature (a non-breaking change which adds functionality)

### DIFF
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -68,7 +68,7 @@ function getValidTemplate(page: string, pages: Pages) {
 
 export function pathFromPageName(page: string, pages: Pages, params: any) {
   const validTemplate = getValidTemplate(page, pages)
-  if(!validTemplate) return null;
+  if (!validTemplate) return null
   return new RouteParser(validTemplate).reverse(params) || null
 }
 
@@ -108,7 +108,7 @@ function getValidParams(id: string, pages: Pages, path: string, params: any) {
   const validParams = getParams(template, path) as Record<string, any>
   const invalidParams = difference(keys(params), keys(validParams))
 
-  if(!isEmpty(invalidParams)){
+  if (!isEmpty(invalidParams)) {
     console.warn(`The following params are invalid:\n${invalidParams.join('\n')}\n`)
   }
   return validParams
@@ -121,7 +121,7 @@ function getRouteFromPageName(
 ): NavigationRoute | null {
   const path = pathFromPageName(id, pages, params) || ''
   const validParams = getValidParams(id, pages, path, params)
-  return path ? { id, path, params: validParams} : null
+  return path ? { id, path, params: validParams } : null
 }
 
 export function getRouteFromPath(
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -1,31 +1,43 @@
+<<<<<<< HEAD
 import { canUseDOM } from 'exenv'
 import { History, LocationDescriptorObject } from 'history'
 import queryString from 'query-string'
 import * as RouteParser from 'route-parser'
+=======
+import { canUseDOM } from "exenv";
+import { History, LocationDescriptorObject } from "history";
+import queryString from "query-string";
+import * as RouteParser from "route-parser";
+import { keys, difference, isEmpty } from "ramda";
+>>>>>>> Improve params display of warning and refactor getValidParams method
 
-const EMPTY_OBJECT = (Object.freeze && Object.freeze({})) || {}
+const EMPTY_OBJECT = (Object.freeze && Object.freeze({})) || {};
 
 function getScore(path: string) {
-  const catchAll = (path.match(/\*/g) || []).length
-  const catchOne = (path.match(/:/g) || []).length
-  const fixed = (path.match(/\/[\w_-]+/g) || []).length
+  const catchAll = (path.match(/\*/g) || []).length;
+  const catchOne = (path.match(/:/g) || []).length;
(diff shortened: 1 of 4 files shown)


react/utils/pages.ts
```diff
@@ -102,12 +109,20 @@ function getRouteFromPageName(
...
 ): NavigationRoute | null {
-  const path = pathFromPageName(id, pages, params)
-  return path ? { id, path, params } : null
+  const path = pathFromPageName(id, pages, params) || ''
+  const template = getValidTemplate(id, pages) || ''
+  const validParams = getParams(template, path) as Record<string, any>
+  const invalidParams = difference(keys(params), keys(validParams))
+
+  if(!isEmpty(invalidParams)){
+    console.warn(`The following params are invalid:\n${invalidParams}\n`)
```
GustavoMA: use `invalidParams.join('\n')` that will make a better warning message.

react/utils/pages.ts
```diff
@@ -102,12 +109,20 @@ function getRouteFromPageName(
   pages: Pages,
   params: any
 ): NavigationRoute | null {
-  const path = pathFromPageName(id, pages, params)
-  return path ? { id, path, params } : null
+  const path = pathFromPageName(id, pages, params) || ''
+  const template = getValidTemplate(id, pages) || ''
```
luciannojunior: Extract 113 to 115 in another function

react/utils/pages.ts
```diff
@@ -61,9 +63,13 @@ export function pathFromPageName(page: string, pages: Pages, params: any) {
...
     return null
   }
+  return adjustTemplate(template)
+}
 
-  const properTemplate = adjustTemplate(template)
-  return new RouteParser(properTemplate).reverse(params) || null
+export function pathFromPageName(page: string, pages: Pages, params: any) {
+  const validTemplate = getValidTemplate(page, pages)
+  if(!validTemplate) return null;
```
luciannojunior:
```suggestion
  if(!validTemplate) return null
```

react/utils/pages.ts
```diff
@@ -61,9 +63,13 @@ export function pathFromPageName(page: string, pages: Pages, params: any) {
...
+  return new RouteParser(validTemplate).reverse(params) || null
```
luciannojunior: The `|| null` might be unnecessary here. If there's a truthy `validTemplate`, enforced in the line above, I think that there's no way the `reverse` will return falsy

luciannojunior:
Careful with merge conflicts

react/utils/pages.ts
fanny: But this check is just for validTemplate, we still can have errors in params, then null could be necessary isn't?

react/utils/pages.ts
GustavoMA: Actually
```suggestion
  if (!validTemplate) return null
```

react/utils/pages.ts
```diff
@@ -97,13 +103,25 @@ function getPagePath(name: string, pages: Pages) {
   return cname && isHost(cname) ? '/' : pagePath
 }
 
+function getValidParams(id: string, pages: Pages, path: string, params: any) {
+  const template = getValidTemplate(id, pages) || ''
+  const validParams = getParams(template, path) as Record<string, any>
+  const invalidParams = difference(keys(params), keys(validParams))
+
+  if(!isEmpty(invalidParams)){
```
GustavoMA:
```suggestion
  if (!isEmpty(invalidParams)) {
```

react/utils/pages.ts
```diff
@@ -97,13 +103,25 @@ function getPagePath(name: string, pages: Pages) {
   return cname && isHost(cname) ? '/' : pagePath
 }
 
+function getValidParams(id: string, pages: Pages, path: string, params: any) {
+  const template = getValidTemplate(id, pages) || ''
+  const validParams = getParams(template, path) as Record<string, any>
+  const invalidParams = difference(keys(params), keys(validParams))
+
+  if (!isEmpty(invalidParams)) {
+    console.warn(`The following params are invalid:\n${invalidParams.join('\n')}\n`)
```
GustavoMA:
```suggestion
    console.warn(`The following params are invalid:\n${invalidParams.join('\n')}`)
```

GustavoMA: No need for this last breakline :)

react/utils/pages.ts
GustavoMA: @luciannojunior `new RouteParser(validTemplate).reverse(params)` can return `false` as well.
This method was already with `|| null` before this changes.

luciannojunior:
image
 The warning is not working (or shouldn't it work in this case? 🤔 ). Take a look

Nice job so far

fanny:
Hmmm, it's seems that we run of different ways, when i run `$r.navigate({page: 'store.product', params: { slug: 'shirt', nome: 'shirt'}})` the warns are showed

vtex-error

but when i run your commag, i get this error:

vtex-e-rror

fanny:
My `$r` is the `RenderProvider` component

PR #290, opened 2019-05-09
Fix to param of navigate object

fanny:

as title says.

When passing query in `to` path, the query object is empty, and search field needs of this values

1. [Acess this workspace](https://fanny--storecomponents.myvtex.com)
2. Create a navigate object passing more params then necessary e.g `navigate({to: '/classic-shoes/p?map=ft'})`
3. And check if `query` object contains `map` field.

#### Types of changes

* [X] Bug fix (a non-breaking change which fixes an issue)
* [ ] New feature (a non-breaking change which adds functionality)

### DIFF
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -2,7 +2,7 @@ import { canUseDOM } from 'exenv'
 import { History, LocationDescriptorObject } from 'history'
 import queryString from 'query-string'
 import * as RouteParser from 'route-parser'
-import { keys, difference, isEmpty } from 'ramda'
+import { difference, is, isEmpty, keys } from 'ramda'
 
 const EMPTY_OBJECT = (Object.freeze && Object.freeze({})) || {}
 
@@ -26,7 +26,11 @@ function trimEndingSlash(token: string) {
 
 function createLocationDescriptor(
   navigationRoute: NavigationRoute,
-  { query, scrollOptions, fetchPage }: Pick<NavigateOptions, 'query' | 'scrollOptions' | 'fetchPage'>
+  {
+    query,
+    scrollOptions,
+    fetchPage,
+  }: Pick<NavigateOptions, 'query' | 'scrollOptions' | 'fetchPage'>
 ): LocationDescriptorObject {
   return {
     pathname: navigationRoute.path!,
@@ -68,7 +72,9 @@ function getValidTemplate(page: string, pages: Pages) {
 
 export function pathFromPageName(page: string, pages: Pages, params: any) {
   const validTemplate = getValidTemplate(page, pages)
-  if (!validTemplate) return null
+  if (!validTemplate) {
+    return null
+  }
   return new RouteParser(validTemplate).reverse(params) || null
 }
 
@@ -79,7 +85,7 @@ export function queryStringToMap(query: string): Record<string, any> {
   return queryString.parse(query)
 }
 
- export function mapToQueryString(query: Record<string, any> = {}): string {
+export function mapToQueryString(query: Record<string, any> = {}): string {
   return queryString.stringify(query)
 }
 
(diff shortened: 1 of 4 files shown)


react/utils/pages.ts
```diff
@@ -154,29 +154,36 @@ export function navigate(
   const {
     page,
     params,
-    query,
-    to,
+    query: inputQuery,
+    to: inputTo = '',
```
Klynger: Is this default value really neccessary? Because you're setting `inputTo = ''` but later you verify if the value is falsy

Klynger: Oh, I see, this string is used later to split the values, but is it possible that this variable can receive `null`?

react/utils/pages.ts
```diff
@@ -154,29 +154,36 @@ export function navigate(
...
     console.error(
       `Invalid navigation options. You should use 'page' or 'to' parameters`
     )
     return false
   }
+  
+  const [to, extractedQuery] = inputTo.split('?')
+  const query = inputQuery || extractedQuery
+
+  if (inputTo && inputQuery) {
```
lbebber: ⛏ ⛏ ⛏ this warning block should be up there, close to the error above, as it doesn't depend on anything else happening

react/utils/pages.ts
```diff
@@ -154,29 +154,36 @@ export function navigate(
...
 
-  if (!page && !to) {
+  if (!page && !inputTo) {
     console.error(
       `Invalid navigation options. You should use 'page' or 'to' parameters`
     )
     return false
   }
+  
+  const [to, extractedQuery] = inputTo.split('?')
```
lbebber: :pick: you could check if the type of `inputTo` is `string` before splitting, and throw an error/log it otherwise.

react/utils/pages.ts
```diff
@@ -154,29 +154,36 @@ export function navigate(
...
+  const query = inputQuery || extractedQuery
+
+  if (inputTo && inputQuery) {
+    console.warn(`You shouldn't pass 'query' in a separate prop when using 'to'`)
```
lbebber: actually, why not? I think there should be a warning only if the user is passing both a `query` and a query string on `to`. otherwise, there's no harm, is there?

luciannojunior: I've discussed with @brunoabreu about this and it's more concise defining only one way to pass queries. Since when using _page_, there's no way to pass the queries there, you need another way, but when using _to_, everything should be passed there.

We are still supporting, but not encouraging

Klynger: @lbebber isn't exatly that what this if statement is doing? :thinking:

react/utils/pages.ts
luciannojunior: Yes, than it will return on the first check

react/utils/pages.ts
luciannojunior: She has set a default to empty string up there. And there's typings for the parameters as well

react/utils/pages.ts
lbebber: ah, I misread it. carry on 👌

react/utils/pages.ts
lbebber: @luciannojunior I meant for the case where a non-string value is sent (rather than no value at all).

lbebber: and typings are currently only strict inside the app, right? there's nothing preventing a non-string value from being sent from another app, if this function is exported, is there?

lbebber: (this is just a minor suggestion btw, I'm for doing these verifications, but it shouldn't really be a problem, you can just ignore this)

luciannojunior: Yes, it makes sense! I will check that. Thanks for the great review

PR #296, opened 2019-05-20
Feature - Add messages

alcar:
#### Added

- Method to add messages to Runtime (`addMessages`).
- `addMessages` and `messages` to Render context.

#### Changed

- `sendInfoFromIframe` signature.
- `updateExtension` and `updateRuntime` now await for `setState` to be finished.

#### Deprecated

- `RenderProvider`'s `updateMessages` private method.
- `sendInfoFromIframe`'s 4th argument (`setMessages`).

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -684,7 +684,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     }
   }
 
-  public updateRuntime = (options?: PageContextOptions) => {
+  public updateRuntime = async (options?: PageContextOptions) => {
     const {
       runtime: { renderMajor },
     } = this.props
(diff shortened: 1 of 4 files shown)


react/components/RenderProvider.tsx
```diff
@@ -782,28 +795,45 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
-    )
+      )
+    })
   }
 
   public handleSetDevice = (device: ConfigurationDevice) => {
     this.setState({ device })
   }
 
+  public addMessages = (newMessages: RenderRuntime['messages']) => {
```
brunoabreu: Should this be sync regardless triggering a promise? Don't we care about this promise result?

react/components/RenderProvider.tsx
```diff
@@ -724,9 +732,14 @@ class RenderProvider extends Component<Props, RenderProviderState> {
           pages,
           route,
           settings,
-        })
-      }
-    )
+        },
+        async () => {
```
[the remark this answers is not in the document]
alcar: I believe it works with async functions, but your suggestion does look more readable to me.

react/components/RenderProvider.tsx
alcar: Not now, but we may soon. I've made it async just in case.

alcar:
@brunoabreu Good point! I've checked with them, and they don't seem to be using it 👍

PR #303, opened 2019-05-29
Fix/preview that respects children

lbebber:
Show preview while the child blocks are being loaded.

Brings back the non-resetting of components as well.

https://lbebber--ecowaterqa.myvtex.com/water-softeners


lbebber:
this PR has been made obsolete

PR #313, opened 2019-06-11
Set Lazy Cookie

tergol1:
Set `VtexWorksapce` cookie.

### DIFF
--- a/react/core/main.tsx
+++ b/react/core/main.tsx
@@ -134,15 +134,15 @@ const render = (
 
   return canUseDOM
     ? ((disableSSR || created
-        ? renderDOM<HTMLDivElement>(root, elem)
-        : hydrate(root, elem)) as Element)
+      ? renderDOM<HTMLDivElement>(root, elem)
+      : hydrate(root, elem)) as Element)
     : renderToStringWithData(root).then(({ markup, renderTimeMetric }) => ({
-        markups: getMarkups(name, markup),
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        maxAge: cacheControl!.maxAge,
-        page,
-        renderTimeMetric,
-      }))
+      markups: getMarkups(name, markup),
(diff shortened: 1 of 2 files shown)


react/core/main.tsx
```diff
@@ -155,6 +155,12 @@ function validateRootComponent(rootName: string, extensions: Extensions) {
   }
 }
 
+function setLazyCookie(setCookie: string) {
+  if (setCookie !== null) {
+    document.cookie = setCookie
```
victorges: Is this really setting the raw value of the cookies? I thought the browser would have some API for accessing the cookies already parsed only.

Also, here you are *setting* the whole cookies, while I believe we should only *append* the lazy cookies to the existing ones. In that case, it'd probably be necessary to check if the cookie we are setting is already set.

react/core/main.tsx
```diff
@@ -155,6 +155,12 @@ function validateRootComponent(rootName: string, extensions: Extensions) {
   }
 }
 
+function setLazyCookie(setCookie: string) {
+  if (setCookie !== null) {
```
victorges: You can check simply `if (setCookie)` which will also be false for `undefined`

react/core/main.tsx
tergol1: Just for the last part, I think that this is not necessary as router won't set the header `X-Vtex-Lazy-Cookie` if it checks the existence of the cookie - we do the "cookie-stuff" inside `FindWorkspaceFromCookies`, where we first check if we can get the Workspace from cookies and then immediately returns the value.

tergol1: I don't know if I understood everything bothering you but I think that is everything ok. Just take a look here 
https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie

Reading it I concluded that the cookie is being set correctly and we don't have to actually care about the other ones that were set before (even after).

firstdoit: Yep. @victorges go back to your golang cave 😘 hahahahaha 😬 😂

PR #316, opened 2019-06-12
Fix Session setting state when unmounted

klzns:


react/components/Session.tsx
```diff
@@ -1,63 +1,64 @@
...
-  private onUpdate() {
-    const {
-      runtime: { ensureSession },
-    } = this.props
-    const { ensured, error } = this.state
+  return [state, safeSetState]
+}
+
+const Session: FunctionComponent = ({ children }) => {
+  const [ensured, setEnsured] = useSafeState(false)
```
lucasecdb: I'd rather do something like [this](https://twitter.com/ryanflorence/status/1129487337375690752) instead of creating this custom hook, although I don't agree much with the naming :joy:

klzns: LOVE IT

react/components/Session.tsx
```diff
@@ -1,63 +1,55 @@
...
-    return <Loading />
-  }
-
-  private onUpdate() {
-    const {
-      runtime: { ensureSession },
-    } = this.props
-    const { ensured, error } = this.state
+const Session: FunctionComponent = ({ children }) => {
+  const [ensured, setEnsured] = useState(false)
```
lbebber: :pick: `isEnsured` talvez?

PR #329, opened 2019-06-26
Avoid recreating React elements when block has composition children

klzns:
Before | After
---|---
image | image

https://breno--alssports.myvtex.com/billab-crossfire-x-short/p?skuId=400192


react/utils/client/links/versionSplitterLink.ts
```diff
@@ -187,7 +187,7 @@ const observableFromOperations = (operations: Operation[], forward: NextLink) =>
   }) as Observable<FetchResult<{ [key: string]: any }, Record<string, any>, Record<string, any>>>
 
 export const versionSplitterLink = new ApolloLink(
-  (operation: Operation, forward?: NextLink) => {
+  (operation: Operation, forward?: NextLink) : any => {
```
jgfidelis: I think this is can be deleted

klzns: It throughs an error if I keep it
image

PR #333, opened 2019-06-27
Add useOwnerBlockCssHandles hook

klzns:
Make it possible to create a CSS handle without creating a CSS file.

Example:

Instead of:

```css
/* style.css */
.container {}
```

```jsx
/* Foo.js */
import React from 'react'
import style from './style.css'

export default function Foo() {
  return <div className={style.container}>Hello</div>
}
```

One would write:

```jsx
/* Foo.js */
import React from 'react'
import { useOwnerBlockCssHandles } from 'vtex.render-runtime' 

const handles = ['container']

export default function Foo() {
  const style = useOwnerBlockCssHandles(handles)

  return <div className={style.container}>Hello</div>
}
```

We should be careful not to write:

```diff
  /* Foo.js */
  import React from 'react'
  import { useOwnerBlockCssHandles } from 'vtex.render-runtime' 
 
- const handles = ['container']
  export default function Foo() {
+   const handles = ['container']
    const style = useOwnerBlockCssHandles(handles)
  
    return <div className={style.container}>Hello</div>
  }
```

That way the `useOwnerBlockCssHandles` would have to compute the handles on every render since the array it's always new.

Two ways to avoid this:

1. (Preferred)
```jsx
const handles = ['container']

export default function Foo() {
  const style = useOwnerBlockCssHandles(handles)
```

2.
```jsx
export default function Foo() {
  const handles = useMemo(() => ['container'], [])
  const style = useOwnerBlockCssHandles(handles)
```

The second option it's worse. It would only make sense if you need to vary the CSS handles based on some prop, that is, filling the second param of `useMemo`.

---

### Important: one behavior it's different from the current CSS file approach!

`useOwnerBlockCssHandles` adds the namespace of the nearest parent block, that means that it will not always add the namespace of the _app_ the component is present.

Example:

- **Foo** is a block and a component from **App A**.
- **Foo** depends on **App B**.
- **App B** have a component **Bar**, which uses `useOwnerBlockCssHandles` with a CSS handle called `item` and have a CSS Module with a handle called `element`.
- **Foo** uses the React component **Bar** internally. 

Render tree
```jsx
<ExtensionPoint id="foo">
  <Foo>
    <Bar />
  </Foo>
</ExtensionPoint>
```

App A
```jsx
/* Foo */
import React from 'react'
import { Bar } from 'vtex.app-b' 

export default function Foo() {
  return <Bar />
}
```

App B
```css
/* style.css */
.element {}
```

```jsx
/* Bar.js */
import React from 'react'
import { useOwnerBlockCssHandles } from 'vtex.render-runtime'
// CSS Modules handles
import style from './style.css'

const handleNames = ['item']

export default function Bar() {
  // CSS handles by hooks
  const handles = useOwnerBlockCssHandles(handleNames)

  return <div className={`${handles.item} ${style.element}`}>Hello</div>
}
```

Rendered HTML

```html
<div class="vtex-app-a-1-x-item vtex-app-b-1-x-element>Hello</div>
```

### Why don't we keep the same behavior as CSS Modules?

We think the CSS handles from a dependency shouldn't be named after their app name, that's an implementation detail. This makes a series of inconveniences, for example:

A user wants to customize the CSS of the price element a shelf block, the user have to use a `:global` scape hatch:

```css
/* vtex.shelf.css */
.container :global(.vtex-store-components-1-x-product-price) {
  margin-top: 10px;
}
```

If the product-price used the hook it would look like:

```css
/* vtex.shelf.css */
.container .product-price {
  margin-top: 10px;
}
```

--- 

One thing to keep in mind with that change is: since all handles could be exported to a parent component, to avoid clashing with the parent component handles it could be good pratice to give them more descriptive names, that is:
```diff
- .container
+ .productPrice-container
```

So if shelf wants to use the product-price it wouldn't have to mind clashing with other classes of the other app.

Example:

- App store-components defines a handle `.container`
- App shelf uses store-components
- If Shelf defined a handle also called `.container` customizations to the Shelf `.container` would also hit store-components `.container`.

We could change this PR to add a namespace parameter so it would do something like: `product-price` + handle

### DIFF
--- a/react/components/ExtensionPoint.tsx
+++ b/react/components/ExtensionPoint.tsx
@@ -7,7 +7,7 @@ import { getImplementation } from '../utils/assets'
 import { TreePathContext, TreePathProps, withTreePath } from '../utils/treePath'
 
 import ExtensionPointComponent from './ExtensionPointComponent'
-import { HostAppProvider } from './HostApp'
+import { OwnerExtensionProvider } from './OwnerExtension'
 import Loading from './Loading'
 import { RenderContext } from './RenderContext'
 import TrackEventsWrapper from './TrackEventsWrapper'
@@ -146,7 +146,7 @@ class ExtensionPoint extends Component<ExtendedProps, State> {
       props,
       <TrackEventsWrapper events={track} id={id}>
         <TreePathContext.Provider value={{ treePath: newTreePath }}>
-          <HostAppProvider>
+          <OwnerExtensionProvider>
             {component ? (
               <ExtensionPointComponent
                 component={component}
@@ -159,7 +159,7 @@ class ExtensionPoint extends Component<ExtendedProps, State> {
             ) : (
               <Loading />
             )}
-          </HostAppProvider>
+          </OwnerExtensionProvider>
         </TreePathContext.Provider>
       </TrackEventsWrapper>
     )
--- a/react/components/ExtensionPoint.tsx
+++ b/react/components/ExtensionPoint.tsx
@@ -7,7 +7,7 @@ import { getImplementation } from '../utils/assets'
 import { TreePathContext, TreePathProps, withTreePath } from '../utils/treePath'
 
 import ExtensionPointComponent from './ExtensionPointComponent'
-import { OwnerExtensionProvider } from './OwnerExtension'
+import { OwnerBlockProvider } from './OwnerBlock'
 import Loading from './Loading'
 import { RenderContext } from './RenderContext'
 import TrackEventsWrapper from './TrackEventsWrapper'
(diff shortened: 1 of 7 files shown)


react/components/ExtensionPoint.tsx
```diff
@@ -7,6 +7,7 @@ import { getImplementation } from '../utils/assets'
 import { TreePathContext, TreePathProps, withTreePath } from '../utils/treePath'
 
 import ExtensionPointComponent from './ExtensionPointComponent'
+import { OwnerBlockProvider } from './OwnerBlock'
```
salesfelipe: Changelog?

react/utils/useCssHandles.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * declarations.
+ * Receives an array of strings ('foo', 'bar') and returns an object with
+ * { foo: 'vendor-appname-1-x-foo', bar: 'vendor-appname-1-x-bar' }.
+ */
+export const useOwnerBlockCssHandles = (handles: string[] = []): CssHandles => {
+  const { identifier } = useOwnerBlock()
+
+  const values = useMemo(() => {
+    if (!identifier) {
+      const fallbackValues = useMemo(() => {
```
lbebber: you shouldn't call a hook inside a condition. either remove the memo or move it up in the scope

PR #343, opened 2019-07-09
useDevice hook

lbebber:
Creates a `useDevice` hook (along with a `<DeviceDetector>` component) for consistent device detection across both SSR and CSR.

Example of usage can be seen here: https://github.com/vtex-apps/store-header/pull/98/commits/fed84f93e78efc812aedf8ec49b3ef12f198d742

Also adds a `useSSR` hook for SSR detection.

Also uses internal code for the SSR detection, removing the `react-no-ssr` package.


lbebber:
Closed in favour of https://github.com/vtex-apps/render-runtime/pull/344

lbebber:
(Will be a separate package tho)

PR #345, opened 2019-07-10
Get block title when mounting extensions

thiagomurakami:
Pass `block.title` to `generateExtension` function result.

[ch14958]


felippenardi:
@thiagomurakami Can you add a user-focused description so Bia can write the release notes?

PR #348, opened 2019-07-11
Leaner react tree 🌲 by using hooks

thiagomurakami:
## **Is your release a breaking change, new feature, deprecation or an improvement?**

Improvement.

## **What is it about?** (w/ media if possible)

Improve React render performance. This improvement is more noticeable in pages that have lots of extensions.

#### Before

react_dev_before

#### After

react_dev_after

#### Before

timeline_before

#### After

timeline_after

------

It can lead to improvements of up to 30%, the average is around 20%.

- **What changed?**

Reduce the number of components that React has to mount by using hooks.
Before:

tree_before

After:

tree_after

## **What advantages does it bring the user?**

Better responsiveness in stores and admins.

## **What problem was solved?**

🐌🐌🐌🐌🐌

## **Why should it be used?**

Because it's faster 🏃‍♂️💨

## **How do you implement it?**

🙏 and wait for house keep.

## **What are its collateral effects?**

If I did this right, there should be none. However, it could break everything 😉


manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "render-runtime",
-  "version": "8.41.0",
+  "version": "8.41.1-hkignore.0",
```
felippenardi: do you have to change it?

react/components/RenderProvider.tsx
```diff
@@ -159,7 +153,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
 
   public sendInfoFromIframe = debounce(
     (params?: { shouldUpdateRuntime?: boolean }) => {
-      if (!isStorefrontIframe) {
+      if (!isSiteEditorIframe) {
```
felippenardi: ♥️ 👀

vitoriaHeliane:
:clap: :100:

react/components/ExtensionPointComponent.tsx
```diff
@@ -213,12 +219,51 @@ class ExtensionPointComponent extends PureComponent<
       delete props.__clearError
     }
 
-    return Component ? (
-      <Component {...props}>{children}</Component>
-    ) : (
-      children || <Loading />
+    return (
+      <TreePathContext.Provider value={{ treePath }}>
```
klzns: Use memo here

```tsx
const treePathValue = useMemo(() => ({ treePath }), [treePath])

<TreePathContext.Provider value={treePathValue}>
```

react/components/RenderPage.tsx
```diff
@@ -1,39 +1,36 @@
...
+      nestedPage={page}
+      query={query}
+      params={params}
+      runtime={runtime}
+    >
+      <ExtensionPoint
+        id={page}
+        query={query}
+        params={params}
+        treePath=""
```
klzns: Is that right?

react/core/main.tsx
```diff
@@ -129,7 +129,7 @@ const render = (
       root={name}
       runtime={runtime}
     >
-      {!isPage ? <ExtensionPoint id={name} /> : null}
+      {!isPage ? <ExtensionPoint id={name} treePath="" /> : null}
```
klzns: why is that? `treePath=""`

react/components/ExtensionPointComponent.tsx
thiagomurakami: This is a class component 😬 
I can make a simple cache logic.

klzns: You can use https://www.npmjs.com/package/memoize-one for that

klzns: You can also create a memoized provider

```tsx
`const TreePathContextProvider = memo(({ treePath, children }) =>
  <TreePathContext.Provider value={{ treePath }}>
    {children}
  </TreePathContext.Provider>
)
```

thiagomurakami:
Hot reload broke :(

PR #351, opened 2019-07-11
[Second attempt] Leaner react tree 🌲 by using hooks

thiagomurakami:
## **Is your release a breaking change, new feature, deprecation or an improvement?**

Improvement.

## **What is it about?** (w/ media if possible)

Improve React render performance. This improvement is more noticeable in pages that have lots of extensions.

#### Before

react_dev_before

#### After

react_dev_after

#### Before

timeline_before

#### After

timeline_after

------

It can lead to improvements of up to 30%, the average is around 20%.

- **What changed?**

Reduce the number of components that React has to mount by using hooks.
Before:

tree_before

After:

tree_after

## **What advantages does it bring the user?**

Better responsiveness in stores and admins.

## **What problem was solved?**

🐌🐌🐌🐌🐌

## **Why should it be used?**

Because it's faster 🏃‍♂️💨

## **How do you implement it?**

🙏 and wait for house keep.

## **What are its collateral effects?**

If I did this right, there should be none. However, it could break everything 😉


brunoabreu:
:aqui-tem-coragem:

PR #358, opened 2019-07-23
Change href for links inside admin iframe

gustavorosolem:
In admin builder, all internal links must be `/admin/app/*anything` to proper appear inside of the iframe.
But due to this, if user access the link directly, like when you open in a new tap, he will open the `/admin/app/` url, and will se the admin without the sidebar.

Screen Shot 2019-07-24 at 15 07 43
Screen Shot 2019-07-24 at 15 08 08

With this change, the href show properly, so now user can open in a new tap without worry:

Screen Shot 2019-07-24 at 15 09 52

**How this works**
The `href` attribute is just a fallback for the browser, when the user click in a link, he will call the function on `onClick`, and this one will still use the correct `/admin/app/` pattern.

### DIFF
--- a/react/components/Link.tsx
+++ b/react/components/Link.tsx
@@ -76,7 +76,7 @@ const Link: React.FunctionComponent<Props> = ({
 
   const href = getHref()
   // Href inside admin iframe should omit the `/app/` path
-  const hrefWithoutIframePrefix = domain === 'admin' && href.startsWith('/admin/app/') ? href.replace('/admin/app/', '/admin/') : href
+  const hrefWithoutIframePrefix = domain && domain === 'admin' && href.startsWith('/admin/app/') ? href.replace('/admin/app/', '/admin/') : href
 
   return (
     <a href={hrefWithoutIframePrefix} {...linkProps} onClick={handleClick}>
--- a/react/components/Link.tsx
+++ b/react/components/Link.tsx
@@ -26,7 +26,7 @@ const Link: React.FunctionComponent<Props> = ({
   children,
   ...linkProps
 }) => {
-  const { pages, navigate, rootPath = '' } = useRuntime()
+  const { pages, navigate, rootPath = '', route: { domain } } = useRuntime()
 
   const handleClick = useCallback(
     (event: MouseEvent<HTMLAnchorElement>) => {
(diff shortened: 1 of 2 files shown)


react/components/Link.tsx
```diff
@@ -74,8 +74,12 @@ const Link: React.FunctionComponent<Props> = ({
     return '#'
   }
 
+  const href = getHref()
+  // Href inside admin iframe should omit the `/app/` path
+  const hrefWithoutIframePrefix = domain && domain === 'admin' && href.startsWith('/admin/app/') ? href.replace('/admin/app/', '/admin/') : href
```
kevinch: maybe `domain.toLowerCase()` just to be sure `Admin` won't interfere?

gustavorosolem: That's no need, this domain has just two options (`admin` and `store`) and is not set by the user/developer, but by render.

firstdoit:
Ohh that problem is really nasty! 
Me and @ohmyguigs discussed it, and we concluded this was not the best solution since it will make the runtime "aware" of a layer above it. We decided we would implement a wrapper component on `admin.app` that, upon rendering, if it notices it is *not* in an iframe and the path starts with `/admin/app/*`, then it redirects to `/admin/*`.
How does that sound, @gustavorosolem ?

klzns:
@firstdoit I think we could merge this PR and revert it after the final solution.

gustavorosolem:
> Ohh that problem is really nasty!
> Me and @ohmyguigs discussed it, and we concluded this was not the best solution since it will make the runtime "aware" of a layer above it. We decided we would implement a wrapper component on `admin.app` that, upon rendering, if it notices it is _not_ in an iframe and the path starts with `/admin/app/*`, then it redirects to `/admin/*`.
> How does that sound, @gustavorosolem ?

@firstdoit I think it is a bit ugly, aesthetically speaking, because user will see the `/app/` when putting the mouse over the link, or will copy the url with the `/app/` if he right click and select to copy path.
But I understand yours reasons, and yes, can be done with a redirect.
Just to alignment, you pretend to create this redirect logic in `vtex.admin` app or some other global place? If is in `vtex.admin`, I will start to do in the `gocommerce.admin` right now.

klzns:
I also think the redirect would be a bit weird. I would rather change the href more elegantly or they way it is right now, until it becomes extra weird.

gustavorosolem:
In GoCommerce, for better UX, we use the `<Link>` in everywhere, like in tables, navigation, breadcrumbs, buttons. One simple click in the list of orders is a `<Link>`.

Is know, that users in daily routines, open orders using *new tab*, to optimize their time and manage multiple orders at same time.

Due to the lack of interactions in this matter and since it had already been approved, I will merge, and in the future, if we found a better solution, we can change.

PR #359, opened 2019-07-23
Feature separate admin lang from store lang

Jeymisson:
It will change admin's language without changing store's language.
It should only be merged after this one: https://github.com/vtex/node-vtex-api/pull/190

Test: https://jey--storecomponents.myvtex.com/admin/cms/storefront
1 - Select an admin language.
2 - At storefront's topbar, select a different language for the store.
3 - Store language now should be different from admin's language.

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -694,21 +694,28 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     return assetsPromise
   }
 
-  public onLocaleSelected = (locale: string) => {
+  public onLocaleSelected = (locale: string, domain?: string) => {
     if (locale !== this.state.culture.locale) {
-      const sessionData = {
-        public: {
-          cultureInfo: {
-            value: locale,
+      const sessionData = { public: { } }
+      if(domain && domain === 'admin'){
+        sessionData.public = {
+            admin_cultureInfo: {
+              value: locale,
           },
-        },
+        }
+      } else {
+        sessionData.public = {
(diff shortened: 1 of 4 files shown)


react/components/RenderProvider.tsx
```diff
@@ -692,21 +692,28 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
+  public onLocaleSelected = (locale: string, domain?: string) => {
     if (locale !== this.state.culture.locale) {
-      const sessionData = {
-        public: {
-          cultureInfo: {
-            value: locale,
+      const sessionData = { public: { } }
+      if(domain && domain === 'admin'){
+        sessionData.public = {
+            admin_cultureInfo: {
```
thiagomurakami: Maybe DRY on this and declare `cultureInfo` to be reused?
```
const cultureInfo = { 
  value: locale,
}
```

react/utils/dom.tsx
```diff
@@ -83,7 +83,8 @@ export const isSiteEditorIframe =
   canUseDOM &&
   window.top !== window.self &&
   window.top.__provideRuntime &&
-  !!window.top.__provideRuntime
+  !!window.top.__provideRuntime &&
+  window.__RUNTIME__.route.domain === 'store'
```
thiagomurakami: 🙏

react/components/RenderProvider.tsx
Jeymisson: Session expects it like that. We might improve this later.

PR #364, opened 2019-08-04
Fix/error boundaries

lbebber:
Adds error boundaries on the ExtensionPoint component, limiting crashes to the component instead of breaking most of the page.
Also hides errors on production mode.

Before (simulating a crash on the Shelf component): https://lbebber--storecomponents.myvtex.com/?__disableSSR
After: https://lbebber2--storecomponents.myvtex.com/?__disableSSR

### DIFF
--- a/react/components/ErrorBoundary.tsx
+++ b/react/components/ErrorBoundary.tsx
@@ -3,7 +3,7 @@ import ExtensionPointError from './ExtensionPointError'
 import { useRuntime } from './RenderContext'
 
 interface Props {
-  runtime: any
+  runtime: RenderContext
 }
 
 class ErrorBoundary extends React.Component<Props> {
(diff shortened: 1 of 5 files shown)


react/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { ErrorInfo, FunctionComponent } from 'react'
+import ExtensionPointError from './ExtensionPointError'
+import { useRuntime } from './RenderContext'
+
+interface Props {
+  runtime: any
```
jgfidelis: there is a `RenderRuntime` type

lbebber: @jgfidelis good catch, left this as a WIP and forgot to update

PR #370, opened 2019-08-08
Add render logic for AMP pages

lucasecdb:


react/core/main.tsx
```diff
@@ -133,17 +134,60 @@ const render = (
...
-        maxAge: cacheControl!.maxAge,
-        page,
-        renderTimeMetric,
-      }))
+  if (canUseDOM) {
+    const renderFn = disableSSR || created ? renderDOM : hydrate
+
+    renderFn(root, elem)
+
+    return (undefined as unknown) as Element
```
klzns: Why not `return renderFn(root, elem)`?

react/core/main.tsx
```diff
@@ -223,7 +267,7 @@ function start() {
       // Expose render promise to global context.
       window.rendered = (maybeRenderPromise as Promise<
         NamedServerRendered
-      >).then(({ markups, maxAge, page, renderTimeMetric }) => ({
+      >).then(({ markups, maxAge, page, renderTimeMetric, ...rendered }) => ({
```
klzns: What are you using from `rendered`?

react/core/main.tsx
lucasecdb: it still warns because `render` and `hydrate` are typed to also return `void`, but I can just do the type assertion `as Element`

react/core/main.tsx
lucasecdb: `ampScripts` and `ampHeaderBoilerplate`

lucasecdb: I find it weird to destructure all properties if I just need to forward them to the return value

react/core/main.tsx
lucasecdb: oh, I need to convert it to `unknown` as well.

PR #371, opened 2019-08-12
Bundle assets

Jeymisson:
Check for assets in bundled urls to fetch only necessary ones.


react/utils/components.ts
```diff
@@ -1,3 +1,5 @@
+import path from 'path'
```
brunoabreu: Let's avoid importing node native modules

PR #373, opened 2019-08-19
Fix lint

Jeymisson:
- Fixed lint errors
- Add husky to prevent commits with lint errors

### DIFF
--- a/react/components/ErrorBoundary.tsx
+++ b/react/components/ErrorBoundary.tsx
@@ -7,32 +7,25 @@ interface Props {
 }
 
 class ErrorBoundary extends React.Component<Props> {
-  state = {
+  public state = {
     error: undefined,
     errorInfo: undefined,
   }
 
-  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+  public componentDidCatch(error: Error, errorInfo: ErrorInfo) {
     this.setState({
       error,
       errorInfo,
     })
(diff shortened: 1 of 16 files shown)


react/components/ErrorBoundary.tsx
```diff
@@ -7,32 +7,33 @@ interface Props {
...
 
 class ErrorBoundary extends React.Component<Props> {
-  state = {
+  private _state = {
     error: undefined,
     errorInfo: undefined,
   }
 
-  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+  public get state() {
```
thiagomurakami: Since it's React API, maybe it isn't a good idea to shouldn't override `this.state`

react/components/ErrorBoundary.tsx
```diff
@@ -7,32 +7,33 @@ interface Props {
 }
 
 class ErrorBoundary extends React.Component<Props> {
-  state = {
+  private _state = {
```
thiagomurakami: Maybe `public state`? Why did you need to use `private _state`?

Jeymisson: It was a vscode suggestion, but believe they're the same thing.

Jeymisson: Will leave with the pattern we already use though.

PR #376, opened 2019-08-22
Delayed prefetch pages

Jeymisson:
- Add a pool to prefetch pages to only execute it once.
- Execute prefetch pages when there is no cpu usage `requestIdleCallback ` or after waiting for 3 secs
- Add querystring to disable prefetchPages: `__disablePrefetchPages`

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -72,6 +72,7 @@ export interface RenderProviderState {
 }
 
 const SEND_INFO_DEBOUNCE_MS = 100
+const DISABLE_PREFETCH_PAGES = '__disablePrefetchPages'
 
 const noop = () => {}
 
@@ -171,6 +172,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
   private sessionPromise: Promise<void>
   private unlisten!: UnregisterCallback | null
   private apolloClient: ApolloClient<NormalizedCacheObject>
+  private prefetchRoutes: Set<string>
 
   public constructor(props: Props) {
     super(props)
@@ -251,6 +253,8 @@ class RenderProvider extends Component<Props, RenderProviderState> {
       route,
       settings,
     }
+
+    this.prefetchRoutes = new Set<string>()
   }
 
   public componentDidMount() {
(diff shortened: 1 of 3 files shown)


react/components/RenderProvider.tsx
```diff
@@ -622,45 +633,18 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
-      messages: { ...defaultMessages, ...this.state.messages, ...messages },
-    }))
+    const { prefetchRoutes } = this.state
+    const disablePrefetch =
+      query &&
+      DISABLE_PREFETCH_PAGES in query &&
+      query[DISABLE_PREFETCH_PAGES] !== 'false'
+    if (!disablePrefetch) {
+      routeIds.forEach(routeId => prefetchRoutes.add(routeId))
+      this.setState({ prefetchRoutes })
```
brunoabreu: I think we should never use `setState` here. Prefetching should be lazy.

`prefetchRoutes` can be just a private field of `RenderProvider`

One concern is that `prefetchDefaultPages` is a function always available in runtime context, but it only works if you call it before RenderProvider is mounted. If that is the behavior we want we should at least log a warning and do nothing if someone calls it after `componentDidMount`.

react/components/RenderProvider.tsx
```diff
@@ -944,6 +928,61 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
+    const {
+      components: defaultComponents,
+      extensions: defaultExtensions,
+      messages: defaultMessages,
+    } = await fetchDefaultPages({
+      apolloClient: this.apolloClient,
+      locale,
+      pages,
+      renderMajor,
+      routeIds: Array.from(prefetchRoutes),
```
brunoabreu: Better clean up prefetchRoutes

```ts
const routeIds = Array.from(prefetchRoutes)
prefetchRoutes.clear()
```

Jeymisson: prefetchRoutes is only read once, its nice to have it's data in case we need debugging.

react/components/RenderProvider.tsx
Jeymisson: Good point! I think we should only execute prefetchPages once, it doesn't make sense to call it after RenderProvider is already rendered.
I'm throwing an exception when someone calls it after render provider have been already rendered.

PR #405, opened 2019-09-12
Separate locale by product

Jeymisson:


react/utils/client/index.ts
```diff
@@ -53,7 +53,8 @@ export const getState = (runtime: RenderRuntime) => {
 }
 
 export const getClient = (runtime: RenderRuntime, baseURI: string, runtimeContextLink: ApolloLink, ensureSessionLink: ApolloLink, cacheControl?: PageCacheControl) => {
-  const {account, workspace} = runtime
+  const {account, workspace, culture: { locale }, route } = runtime
+  const domain = route.path && route.path.split('/')[1]
```
brunoabreu: I think it should be `route.id` instead of `route.path`

Screen Shot 2019-09-16 at 15 06 52

PR #414, opened 2019-09-24
Use disableSSQ for AMP pages

lucasecdb:
this PR updates the AMP page rendering to also disable server-side queries with the `disableSSQ` flag. I also moved some eslint configurations to the root folder, as it wasn't working correctly (at least in my editor).

I added some logic for the component lazy-load in the `ExtensionPoint` component, because we can't lazy load components in AMP pages, but if anyone feels a bit itchy about it I can remove that (right now I don't think it's working 100%, I still need to remove the `"render": "client"` on the Shelf app for it to render, if someone has any idea about it I'd highly appreciate hearing it).

### DIFF
--- a/react/components/ExtensionPoint.tsx
+++ b/react/components/ExtensionPoint.tsx
@@ -192,9 +192,12 @@ const ExtensionPoint: FC<Props> = props => {
     </ExtensionPointComponent>
   )
 
-  // `client` component assets are sent to server side rendering, but they should display a loading animation.
-  // `lazy` components might never be used, so they don't necessarily need a loading animation.
-  return renderStrategy === 'client' ? (
+  // "client" component assets are sent to server side rendering,
+  // but they should display a loading animation.
+  //
+  // "lazy" components might never be used, so they don't necessarily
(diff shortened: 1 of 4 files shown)


react/components/ExtensionPoint.tsx
```diff
@@ -192,9 +192,12 @@ const ExtensionPoint: FC<Props> = props => {
...
 
-  // `client` component assets are sent to server side rendering, but they should display a loading animation.
-  // `lazy` components might never be used, so they don't necessarily need a loading animation.
-  return renderStrategy === 'client' ? (
+  // "client" component assets are sent to server side rendering,
+  // but they should display a loading animation.
+  //
+  // "lazy" components might never be used, so they don't necessarily
+  // need a loading animation.
+  return renderStrategy === 'client' && !runtime.amp ? (
```
brunoabreu: The `client` blocks become `clientOnly` assets in `render-server`. Those assets are not sent to `render-ssr`, so it won't work as you expect unless you do some changes in `render-server`.

PR #416, opened 2019-09-25
Add support for composition list

lbebber:


lbebber:
we went on another direction

PR #418, opened 2019-09-25
Improves navigation preview

vwraposo:
Previews navigation page if it is known.


react/components/RenderProvider.tsx
```diff
@@ -525,15 +556,49 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
+          contentMap!,
+          pagesState[page]
+        )
+      }
+      this.setState(
+        {
+          extensions: replaceExtensionsWithDefault(
+            { ...updatedExtensions, ...this.state.extensions },
+            page,
+            defaultExtensions
```
brunoabreu: It's always empty. We don't fetch default extensions, we are just prefetching components.

PR #430, opened 2019-10-14
Feat/intl3.x

jgfidelis:
Bump react-intl to major 3.x

https://intl4--storecomponents.myvtex.com/
https://intl--boticario.myvtex.com
https://intl--tbb.myvtex.com/la-vie-est-belle-eau-de-parfum-lancome-perfume-feminino-804902-p/p
https://intl--exitocol.myvtex.com/lavadora-carga-superior-13-kg-29-lb-silver-653335/p
https://intl--samsungar.myvtex.com/50--smart-tv-samsung-4k-ultra-hd/p
https://intl--parquedpedro.myvtex.com/bone-new-era-3930-basic-preto-e-branco-raiders-nfl-286436/p
https://intl--biscoind.myvtex.com/standard-fastenings-3rswpr04008/p
https://intl--motorolauk.myvtex.com/smartphones-motorola-one-zoom/p
https://intl--gc-jjg3536.mygocommerce.com/produtocypress/p
https://intl--vtexhelp.myvtex.com/


react/core/main.tsx
```diff
@@ -115,14 +115,14 @@ const render = async (
...
-    culture: { locale },
   } = runtime
 
   const cacheControl = canUseDOM ? undefined : new PageCacheControl()
   const baseURI = getBaseURI(runtime)
   registerEmitter(runtime, baseURI)
   emitter = runtime.emitter
-  addLocaleData(locale)
+
+  await import('../intl-polyfill')
```
brunoabreu: We should only do it if browser does not implement `Intl.PluralRules` or `Intl.RelativeTimeFormat`. Otherwise we are spending network bandwidth with an asset that is useless.

It should also be done out of here. We don't need to wait for the render function to be called. We want these assets to be fetched as soon as we discover the browser needs a polyfill.

brunoabreu: Also, check if this is running on client-side before doing it.

react/intl-polyfill.ts
```diff
@@ -0,0 +1,33 @@
+import { path } from 'ramda'
+
+const locale = path<string>(['__RUNTIME__', 'culture', 'locale'], window)
+const [lang] = locale ? locale.split('-') : ''
+
+if (lang && window.Intl && !window.Intl.PluralRules) {
+  if (!window.IntlPluralRules || !window.IntlPluralRules.PluralRules) {
+    import('@formatjs/intl-pluralrules/polyfill' as any).then(() => {
```
brunoabreu: I think we could always import both polyfills. We are creating a waterfall of dynamic imports that probably will give us a worse page load with slow connections.

react/intl-polyfill.ts
```diff
@@ -0,0 +1,33 @@
...
+      import('@formatjs/intl-pluralrules/dist/locale-data/' + lang).then(() => {
+        window.Intl.PluralRules = window.IntlPluralRules.PluralRules
```
brunoabreu: There is no `window.IntlPluralRules` when you use `polyfill`. It already put itself directly into `window.Intl.PluralRules`

Not sure why you are not getting some error here. 🤔 
Did you check the console? It would log an unhandled promise exception.

react/utils/reactIntl.tsx
```diff
@@ -0,0 +1,154 @@
...
+  if (!timeDiff) {
+    return null
+  }
+
+  const { value: newValue, unit } = getCorrectUnit(timeDiff)
+
+  return intl.formatRelativeTime(newValue, unit)
+}
+
+const renderRuntimeInjectIntl: FC<any> = (
```
brunoabreu: :kaotico:

Will it have an impact on performance? =/

jgfidelis: I don't think so

PR #432, opened 2019-10-16
Remove unnecessary "?" from path

thiagomurakami:
When there isn't a querystring, the routePath shouldn't have `?` in it.

It was causing some issues in `admin-pages`, because by having `?`, the `updateRuntime` function makes a request without `__pickRuntime` (returning a HTML instead of a JSON).

This PR aims to fix this by relying on the behaviour that `stringify({}) === ''`.


brunoabreu:
Please rebase before merging

PR #444, opened 2019-10-29
Session first

tlgimenes:
Makes the session query in the beginning of the rendering. 
Also, this PR awaits for queries only when fetching private/segment queries

### DIFF
--- a/react/utils/client/links/ensureSessionLink.ts
+++ b/react/utils/client/links/ensureSessionLink.ts
@@ -14,14 +14,8 @@ export const ensureSessionLink = (sessionPromise: Promise<any>) => {
           ? sessionPromise
           : Promise.resolve()
 
-        const rand = Math.random().toFixed(3)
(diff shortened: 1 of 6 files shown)


react/utils/client/links/ensureSessionLink.ts
```diff
@@ -0,0 +1,36 @@
+import { ApolloLink, NextLink, Observable, Operation } from 'apollo-link'
+import { Subscription } from 'apollo-client/util/Observable'
+
+export const ensureSessionLink = (sessionPromise: Promise<any>) => {
+  const sessionRequiredForScope = new Set(['segment', 'private'])
```
jgfidelis: nitpick: this could be defined out of the function.

tlgimenes: It was explicit, since I don't want to generate unwanted objects in SSR. However it could harm our beloved moto g4 ... Maybe I'll change it

PR #447, opened 2019-10-31
Feature/stable loading

lbebber:
Adds:
- Export LoadingContext component, to allow data fetchers display a loading status while the data is being loaded.
- LoadingWrapper to LayoutContainer, which displays a loading status while the components are being loaded.

Changes:
- ExtensionPointComponent content is not rendered while the component is being loaded.
- Uses a loading bar at the top of the page in lieu of an entire-page loader while the layout info is being loaded

Works in tandem with a forthcoming PR on `vtex.store`

Test on: https://lbebber5--alssports.myvtex.com/


react/components/LoadingContext.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const { isParentLoading } = useContext(LoadingContext)
+
+  const contentContainer = useRef<HTMLDivElement>(null)
+  const loaderContainer = useRef<HTMLDivElement>(null)
+
+  const loadingTimeout = useRef<NodeJS.Timer | null>(null)
+  const loadingComplete = useRef(false)
+
+  const updateLoading = useCallback(() => {
+    const areComponentsLoaded = Object.values(state.current.components).some(
```
jgfidelis: is this name correct? Shouldn't it be named `areComponentsLoading`?

lbebber: good catch!

react/components/LoadingContext.tsx
```diff
@@ -0,0 +1,179 @@
...
+  }, [isParentLoading, updateLoading])
+
+  const value = useMemo(
+    () => ({
+      setLoading,
+    }),
+    [setLoading]
+  )
+
+  const isSSR = !window.navigator
```
jgfidelis: I would use canUseDOM here just for consistency

react/components/LoadingContext.tsx
```diff
@@ -0,0 +1,179 @@
...
+export const withLoading = <T extends {}>(Component: ComponentType<T>) => {
+  const EnhancedComponent: FunctionComponent<T> = props => {
+    const isSSR = !window.navigator
+
+    if (isSSR) {
+      return <Component {...props} setLoading={() => {}} />
+    }
+
+    return (
+      <LoadingContext.Consumer>
```
jgfidelis: use `useContext` here

react/components/RenderProvider.tsx
```diff
@@ -962,7 +959,7 @@ class RenderProvider extends Component<Props, RenderProviderState> {
 
   private prefetchPages = () => {
     if (this.prefetchRoutes.size > 0) {
-      setTimeout(this.execPrefetchPages, 20 * 1000)
+      setTimeout(this.execPrefetchPages, 6 * 1000)
```
Klynger: Why use `setTimeout` instead of idle event from window?

lbebber: @Klynger good question, but this is just an adjustment that has hitched a ride on this PR. using the idle event might be good for another PR

PR #449, opened 2019-11-19
Allow multiple queries from render

vwraposo:


react/components/RenderProvider.tsx
```diff
@@ -920,29 +921,29 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
-      )
-    }
+  private hydrateApolloCache = (
+    queryData: Array<{
+      data: string
+      query: string
+      variables: Record<string, any>
+    }>
+  ) => {
+    forEach(({ data, query, variables }) => {
```
NataliaGodot: NIT: Any reason to use `forEach` and not `map`?

PR #451, opened 2019-11-26
Add setup based on toolbelt command

lucasecdb:
This adds the tooling setup based on the `vtex setup` toolbelt command. This ensures that all editors will be able to correctly infer all eslint rules (since eslint 6 loads the plugins based on the location it is installed on, check more [here](https://eslint.org/docs/user-guide/migrating-to-6.0.0#package-loading-simplification)).


lucasecdb:
I don't think we need to release this, because this only affects the ESLint setup, but let me know if anyone thinks otherwise.

PR #455, opened 2019-12-02
Add navigation route modifier hooks for store URL normalisation

Jeymisson:
- Remove store url normalisation logic
- Add navigation route normalisation hook that will be used by store: [https://github.com/vtex-apps/store/tree/feature/navigation-path-normalization]

### DIFF
--- a/react/utils/pages.ts
+++ b/react/utils/pages.ts
@@ -1,17 +1,7 @@
 import { canUseDOM } from 'exenv'
 import { History, LocationDescriptorObject } from 'history'
 import queryString from 'query-string'
-import {
-  contains,
-  difference,
-  is,
-  isEmpty,
-  keys,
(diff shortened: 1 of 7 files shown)


react/utils/pages.ts
```diff
@@ -332,16 +297,18 @@ export function navigate(
...
+    })
+    navigationRoute.path = path || navigationRoute.path
+    query = fixedQuery || query
   }
 
-  navigationRoute.path = pathToLowerCase(navigationRoute.path, query)
-
   if (history) {
-    const nextQuery = mergePersistingQueries(history.location.search, query)
+    const nextQuery = query // mergePersistingQueries(history.location.search, query)
```
lbebber: ⛏ cleanup comment

PR #456, opened 2019-12-03
Notify on interrupted link

vcalasans:
This PR, together with vtex/sse-server#6, makes the user see a "Link Interrupted" pop-up in the browser whenever a link process is interrupted in the current account/workspace.

### DIFF
--- a/react/components/BuildStatus.tsx
+++ b/react/components/BuildStatus.tsx
@@ -5,6 +5,7 @@ interface State {
   animateOut: boolean
   status: string | null
   anchor: 'left' | 'right'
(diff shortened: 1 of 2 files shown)


react/components/BuildStatus.tsx
```diff
@@ -138,12 +160,15 @@ class BuildStatus extends Component<RenderContextProps, State> {
 
     const fail = (
       <p className="ma2">
+        {' '}
```
jgfidelis: erase

vcalasans: Done!

PR #469, opened 2020-01-02
Fix/page remount

augustolazaro:
Related to https://app.clubhouse.io/vtex/story/28702/admin-quando-navega-p%C3%A1gina-recarrega

Clicking in an item of a list perforrm an entire page remount (including queries and stuff). This changes do almost the same as before, but is more clever when rendering the loading bar.

https://augusto--mystore.mygocommerce.com/admin/catalog/products
https://augusto--storecomponents.myvtex.com/admin/apps/

### DIFF
--- a/react/components/ExtensionPoint.tsx
+++ b/react/components/ExtensionPoint.tsx
@@ -198,7 +198,7 @@ const ExtensionPoint: FC<Props> = props => {
   // need a loading animation.
   return renderStrategy === 'client' && !runtime.amp ? (
     <NoSSR onSSR={<Loading />}>
-      {runtime.preview && !runtime.amp && <LoadingBar />}
+      {runtime.preview && isRootTreePath && <LoadingBar />}
       {extensionPointComponent}
     </NoSSR>
   ) : (
--- a/react/components/ExtensionPoint.tsx
+++ b/react/components/ExtensionPoint.tsx
@@ -196,13 +196,18 @@ const ExtensionPoint: FC<Props> = props => {
   //
   // "lazy" components might never be used, so they don't necessarily
   // need a loading animation.
-  return renderStrategy === 'client' && !runtime.amp ? (
-    <NoSSR onSSR={<Loading />}>
+  return (
+    <Fragment>
       {runtime.preview && isRootTreePath && <LoadingBar />}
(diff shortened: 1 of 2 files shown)


react/components/ExtensionPoint.tsx
```diff
@@ -191,22 +191,16 @@ const ExtensionPoint: FC<Props> = props => {
...
-
   // "client" component assets are sent to server side rendering,
   // but they should display a loading animation.
   //
   // "lazy" components might never be used, so they don't necessarily
   // need a loading animation.
   return renderStrategy === 'client' && !runtime.amp ? (
-    <NoSSR onSSR={<Loading />}>{extensionPointComponent}</NoSSR>
+    <NoSSR onSSR={<Loading />}>
+      {runtime.preview && !runtime.amp && <LoadingBar />}
```
jgfidelis: this `!runtime.amp` is not needed here, we already know its true based on the condition above

augustolazaro: Wrong verification, the correct is `runtime.preview && isRootTreePath && <LoadingBar />`

jgfidelis: the `isRootTreePath` variable is not used anymore, only @lbebber can tell us what should happen

react/components/ExtensionPoint.tsx
```diff
@@ -191,22 +191,16 @@ const ExtensionPoint: FC<Props> = props => {
...
-
   // "client" component assets are sent to server side rendering,
   // but they should display a loading animation.
   //
   // "lazy" components might never be used, so they don't necessarily
   // need a loading animation.
   return renderStrategy === 'client' && !runtime.amp ? (
-    <NoSSR onSSR={<Loading />}>{extensionPointComponent}</NoSSR>
+    <NoSSR onSSR={<Loading />}>
+      {runtime.preview && isRootTreePath && <LoadingBar />}
```
lbebber: you are only inserting the bar if the block has `renderStrategy === client`. 

since the bar is only inserted on root blocks, and they basically never have that render strategy, it will likely never render.

it should either be inserted on the `extensionPointComponent` itself, or in a fragment wrapping this ternary

PR #471, opened 2020-01-13
Better binding switch handling

brunoabreu:


react/utils/cookie.ts
```diff
@@ -0,0 +1,16 @@
...
+  key: string,
+  value: string,
+  path = '/',
+  maxAge: number = ONE_YEAR_S
+) => {
+  if (!canUseDOM) {
+    return
+  }
+
+  window.document.cookie = `${key}=${value};path=${path};max-age=${maxAge}`
```
jgfidelis: Is this erasing all the cookies and setting it only to be `vtex_binding_address` or does this just change the vale of the cookie you are setting?

brunoabreu: It just add the new cookie.

brunoabreu: If the cookie is already there, it will change the value.

PR #475, opened 2020-01-27
Allowing the `extensions` argument to be optional on `fetchComponents`

lucis:
Related to [this PR on pages-graphql](https://github.com/vtex/pages-graphql/pull/336).

**Big Picture**: Tech Solutions Internship project by Júlia Rocha to automatically generate documentation for Store Framework components

- **exposes the `fetchComponents`:** For her project to fetch some component implementation, we have to expose the `fetchComponents` functions to the `RenderProvider`, so she can import it from `useRuntime()`.
- **`extensions` argument to be optional:** The `fetchComponents` function has on its interface two arguments: _components_ and _extensions_. We only care about fetching a specific set of components, and those are not coupled with _extensions_, so, if only components are provided, I "mock" an `extensions` object that will match all provided components. I've chosen this approach to avoid entropy on the method implementation.

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -767,12 +767,14 @@ class RenderProvider extends Component<Props, RenderProviderState> {
     extensions?: RenderRuntime['extensions']
   ) => {
     const { runtime } = this.props
+    // In order for only fetching `components`, we create corresponding extensions
+    // for them if they weren't passed
     if (!extensions) {
       const componentsNames = Object.keys(components)
-      extensions = componentsNames.reduce(
-        (acc, component) => ({ ...acc, [component]: { component } }),
-        {}
-      ) as RenderRuntime['extensions']
+      extensions = componentsNames.reduce((acc, component) => {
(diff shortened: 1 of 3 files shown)


react/components/RenderProvider.tsx
```diff
@@ -762,9 +764,16 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
   public fetchComponents = async (
     components: RenderRuntime['components'],
-    extensions: RenderRuntime['extensions']
+    extensions?: RenderRuntime['extensions']
   ) => {
     const { runtime } = this.props
+    if (!extensions) {
+      const componentsNames = Object.keys(components)
+      extensions = componentsNames.reduce(
+        (acc, component) => ({ ...acc, [component]: { component } }),
```
lbebber: ⛏ I'd suggest doing this instead; it's a bit less tidy but it should have better performance and lower memory impact
```suggestion
        (acc, component) => {
          acc[component] = { component }
          return acc
        }
```

PR #480, opened 2020-02-06
Remove ReactDOMServer from client

lbebber:
Lazy-loads ReactDOMServer when necessary, instead of always loading it on the client

Depends on https://github.com/vtex/builder-hub/pull/889

https://domserver--storecomponents.myvtex.com/


lbebber:
Didn't work properly in the end--perhaps something else has it as a dependency, and it became included into the code. Should look into this further in the future

PR #484, opened 2020-02-17
Feature/lazy images fold

lbebber:
Adds:
- `ExperimentalLazyImages` component
- Support for `__fold__.experimentalLazyImages` block.

Relies on https://github.com/vtex-apps/store/pull/424 for the block support

Workspace incoming

### DIFF
--- a/react/core/main.tsx
+++ b/react/core/main.tsx
@@ -15,8 +15,9 @@ import { getDataFromTree } from 'react-apollo'
 import { hydrate, render as renderDOM } from 'react-dom'
 import { Helmet } from 'react-helmet'
 import NoSSR, { useSSR } from '../components/NoSSR'
-import { isEmpty, prop } from 'ramda'
+import { isEmpty } from 'ramda'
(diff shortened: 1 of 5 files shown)


react/core/main.tsx
```diff
@@ -243,13 +244,18 @@ function start() {
...
 
       if (type === 'img') {
-        props.src = optimizeSrcForVtexImg(vtexImgHost, props.src)
-        if (
-          typeof props.src === 'string' &&
-          props.src.startsWith(vtexImgHost)
-        ) {
+        const { src } = props
+        props.src = optimizeSrcForVtexImg(vtexImgHost, src)
+        if (typeof src === 'string' && src.startsWith(vtexImgHost)) {
```
jgfidelis: this is not them same code as above.
You are comparing the old reference of `src`, so inside the if, you will be evaluating the value BEFORE the optimizeSrcForVtexImg call. Is this intended?

lbebber: this was not intended, nice catch

PR #486, opened 2020-02-18
Fix prop target of link not working properly

Klynger:
The  `Link` component wasn't working properly when you try to use it with `target="_blank"`.

To test this you have to follow the steps:
1. Go to the [workspace](https://klynger--storecomponents.myvtex.com/)
2. Click on the `Quickview` text inside the `ProductSummary` to open the modal
3. Click on the `Mais Detalhes >` in the modal

To see the bug [access this workspace](https://linkbug--storecomponents.myvtex.com/) that doesn't have the changes and do the same steps

### DIFF
--- a/react/components/Link.tsx
+++ b/react/components/Link.tsx
@@ -19,6 +19,7 @@ const isMailToUrl = (url: string) => mailToRegex.test(url)
 interface Props extends NavigateOptions {
   onClick?: (event: React.MouseEvent) => void
   className?: string
+  target?: string
 }
 
 const Link: React.FunctionComponent<Props> = ({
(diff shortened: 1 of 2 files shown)


react/components/Link.tsx
```diff
@@ -65,7 +67,8 @@ const Link: React.FunctionComponent<Props> = ({
         replace,
         modifiersOptions,
       }
-      if (navigate(options)) {
+
+      if ((target === '_self' || !target) && navigate(options)) {
```
Klynger: I would like the opinion of you guys about this, because if the user passes `target="_top"`  or `target="_parent"` it might behave in a way that we don't want

lbebber: comment this on the code

PR #489, opened 2020-02-20
Feature/extension point fc

lbebber:
Non-draft implementation of https://github.com/vtex-apps/render-runtime/pull/485

Offers slight performance improvements + reorganizes components.

https://functionextension--alssports.myvtex.com/

PR #497, opened 2020-03-12
Feature/hydrate on view

lbebber:
First version of partial hydration.

Applies partial hydration on blocks that have their `hydration` value set to `on-view`

Also contains the changes from this PR https://github.com/vtex-apps/render-runtime/pull/489

Required PRs:
https://github.com/vtex-apps/render-runtime/pull/497
https://github.com/vtex/render-server/pull/566
https://github.com/vtex/pages-graphql/pull/348
https://github.com/vtex-apps/store/pull/437
https://github.com/vtex/builder-hub/pull/952

Related PRs:
https://github.com/vtex-apps/store-theme/pull/210
https://github.com/vtex-apps/store-footer/pull/105

### How to test:
https://partialhydration--storecomponents.myvtex.com/
Currently only applied on the home.

Open the page with the network Dev Tools tab filtered by JS.
After the page loads, scroll down slowly, and notice that the assets related to the components that appear into view are being loaded.

Navigate around other pages to check that nothing breaks.

Next, open https://partialhydration--storecomponents.myvtex.com?__disableSSR and verify that it works properly


react/components/Hydration/HydrateOnView.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    window?.document?.addEventListener('scroll', handleScroll)
+    /** If somehow the user interacts with the element but `hasBeenViewed` was false,
+     * sets it to true. */
+    element.addEventListener('mouseover', handleMouseOver)
+
+    return () => {
+      window?.document?.removeEventListener('scroll', handleScroll)
+      element.addEventListener('mouseover', handleMouseOver)
```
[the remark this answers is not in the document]
lbebber: 🤦‍♂ good catch

react/hooks/hydration.tsx
```diff
@@ -0,0 +1,16 @@
+import { useRef } from 'react'
+
+export const useDehydratedContent = (hydrationId: string) => {
+  const dehydratedElement = window?.document?.querySelector?.(
```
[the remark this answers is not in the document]
lbebber: thanks for bringing attention to this, but thinking on it I think `useMemo` is a better suited solution for this. agree?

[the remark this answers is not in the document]
jgfidelis: Yeah, in the end it will have similar effect, nic

react/hooks/hydration.tsx
```diff
@@ -0,0 +1,24 @@
...
+/** Gets HTML content from elements rendered on the server, before being
+ * hydrated on the client (and whether they have been SSRendered at all). */
+export const useDehydratedContent = (hydrationId: string) => {
+  /** On both cases below, the use of useRef is to guarantee that only the values
+   * that it has been initialized with is being considered, to guarantee that it's
+   * only taking into account the HTML that's coming from the server, instead of
+   * eventual elements that might be included later.
+   */
+  const dehydratedElement = useRef(
+    window?.document?.querySelector?.(`[data-hydration-id="${hydrationId}"]`)
```
jgfidelis: are you sure this is executed only on the first render? have you tested?

react/index.tsx
```diff
@@ -83,11 +83,20 @@ if (window.ReactIntl) {
...
     const contentLoadedPromise = new Promise(resolve =>
-      document.addEventListener('DOMContentLoaded', resolve)
-    )
-    Promise.all([contentLoadedPromise, intlPolyfillPromise]).then(
-      window.__RENDER_8_RUNTIME__.start
+      window.addEventListener('DOMContentLoaded', resolve)
     )
+    Promise.all([contentLoadedPromise, intlPolyfillPromise]).then(() => {
+      setTimeout(() => {
+        performance.mark('render-start')
```
jgfidelis: is this preformance available in all browsers?

react/hooks/hydration.tsx
jgfidelis: the querySelector is still called on every render.
If you want to avoid that, you might want to use useState with lazy init

`const [dehydratedElement] = useState(
    () => window?.document?.querySelector?.(`[data-hydration-id="${hydrationId}"]`)
)`

jgfidelis: Or do like this:
```
const dehydratedElement = useRef(null)
if (dehydratedElement.current == null) {
  dehydratedElement.current = window?.document?.querySelector?.(`[data-hydration-id="${hydrationId}"]`)
}
```

react/index.tsx
lbebber: 👀

react/hooks/hydration.tsx
lbebber: gonna go with state

react/index.tsx
kaisermann: It seems it does 
image

PR #499, opened 2020-03-20
Fix/missing extension on navigate

lbebber:
Fixes issue with missing extension on navigation on mobile

To test, open https://lbebber--storecomponents.myvtex.com/ on mobile mode, and navigate to a product page (or any page really)

### DIFF
--- a/react/utils/components.ts
+++ b/react/utils/components.ts
@@ -54,6 +54,9 @@ export const traverseExtension = (
   isRoot: boolean
 ): AssetEntry[] => {
   const extension = extensions[extensionId]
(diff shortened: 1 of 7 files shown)


react/utils/components.ts
```diff
@@ -54,6 +54,9 @@ export const traverseExtension = (
   isRoot: boolean
 ): AssetEntry[] => {
   const extension = extensions[extensionId]
+  if (!extension) {
+    return []
```
lbebber: this is the heart of the PR, the other changes are related to the updated Extensions type

PR #503, opened 2020-03-27
Fix base uri when rendered inside iFrame and parent not in IO (7.x)

jgfidelis:


brunoabreu:
It must be done in the runtime 7.x as well

PR #506, opened 2020-03-30
Feature/use lazy images

lbebber:
Exposes `useExperimentalLazyImage` context


react/core/main.tsx
```diff
@@ -318,6 +322,7 @@ export {
   RenderContextConsumer,
   TreePathContextConsumer,
   LazyImages as ExperimentalLazyImages,
+  useLazyImagesContext as useExperimentalLazyImagesContext,
```
klzns:
```suggestion
  useLazyImagesContext,
```

lbebber: @klzns it didn't originally have the `Context` suffix, but I added it because it might be misleading--all it does is to give a context telling whether it should lazy load images or not

lbebber: simply `useLazyImages` might imply that it does some heavy lifting in lazy loading images

lbebber: regarding example, there will be an example in the `InfoCard` component. i don't intend this API to be made public yet though. that's why I added the `Experimental` prefix, btw

klzns: We need to go back to all of those experimental features and decide what we will do with them. It feels like we are creating many experimental things and not coming back to them, not either removing or promoting them.

klzns: Maybe keeping a list of them

lbebber: 💯

PR #509, opened 2020-04-13
implement prefetch context and prefetch on link

jgfidelis:
https://fidelis--boticario.myvtex.com/

Implement prefetch feature

- Create PrefetchContext (Wraps render Provider)
- Make all `Link` components try to prefetch its data
- Do it only after page is loaded
- After all page changes, pause it for a second so navigation can happen and assets be downloaded
- Make prefetch queue empty after all navigations


react/components/RenderProvider.tsx
```diff
@@ -1108,26 +1132,17 @@ class RenderProvider extends Component<Props, RenderProviderState> {
...
-          variables,
-        })
-      } catch (error) {
-        console.warn(
-          `Error writing query from render-server in Apollo's cache`,
-          error
-        )
-      }
-    }, queryData)
+    hydrateApolloCache(
```
lbebber, line 1135: interesting change. has nothing to do with prefetch, has it? (no problem with it, just wondering)

react/error.tsx
```diff
@@ -134,6 +135,10 @@ if (window.__ERROR__) {
       component: ERROR_PAGE_COMPONENT,
       props: {},
     }
+
+    //     const productCache = new LRUCache({ max: 100 })
+    // const searchCache = new LRUCache({ max: 75 })
+    // const otherCache = new LRUCache({ max: 125 })
```
lbebber, line 141:
```suggestion
```

react/hooks/prefetch.ts
```diff
@@ -0,0 +1,266 @@
...
+    validCache,
+    page,
+    client,
+  })
+
+  pathsState[navigationRoute.path] = { fetching: false, page: navigationPage }
+
+  const declarer = pages[navigationPage]?.declarer
+  if (typeof declarer === 'undefined') {
+    // Should not happen, but lets be safe
```
lbebber, line 107: nice

react/hooks/prefetch.ts
```diff
@@ -0,0 +1,266 @@
...
+export const usePrefetchAttempt = ({
+  inView,
+  page,
+  href,
+  options,
+  waitToPrefetch,
+}: UsePrefetchArgs) => {
+  const runtime = useRuntime()
+  const prefetchState = usePrefetch()
+  const client = useApolloClient() as ApolloClientType
```
lbebber, line 191: why does this need the `as ...`?

react/components/RenderProvider.tsx
lbebber: ah, i get it, after reading the code further

react/hooks/prefetch.ts
```diff
@@ -0,0 +1,266 @@
...
+
+interface UsePrefetchArgs {
+  inView: boolean
+  page?: string
+  href: string
+  options: NavigateOptions
+  waitToPrefetch?: number
+}
+
+const getCacheValidData = (path: string, prefecthState: PrefetchState) => {
```
lbebber, line 162: some comment on what is the purpose of this would be good

react/hooks/prefetch.ts
```diff
@@ -0,0 +1,266 @@
...
+      if (validCache.pathValid && validCache.routeValid) {
+        // cache all valid, no need to fetch anything
+        return
+      }
+
+      if (!validCache.pathValid) {
+        pathsState[navigationRoute.path] = { fetching: true }
+      }
+
+      const priority = page === 'store.product' ? 1 : 0
```
lbebber, line 235: this is kind of weird to be hardcoded here inside this function. could this be a separate function? (as in, `getPagePriority(page)`)

react/hooks/prefetch.ts
lbebber: or like, what does "cache valid data" means here.

react/hooks/prefetch.ts
jgfidelis: because typescript

PR #521, opened 2020-05-15
Feature/device blocks

lbebber:
Allows blocks from different devices to load lazily when the screen resizes. 
Incidentally, allows mobile blocks to not load on desktop and vice versa, improving the performance a little bit, and fixes an issue of missing blocks on certain devices/situations/tablet sizes.

### How to test
Open https://deviceblocks--alssports.myvtex.com/

1 - Via devtools, resize the viewport to mobile dimensions, and notice that the header will load lazily.
(there's an unrelated CSS bug that affects the content size in this example)
desktop-mobile

Doing the same on https://www.als.com/ displays the mobile header right away (which means it's already loaded on desktop, which might not be optimized)

2 - via devtools, on mobile mode, refresh the page. It should load the mobile header straight away. Resize it to desktop, and the desktop header should load lazily.

mobile-desktop

Doing the same on https://www.als.com/ should either display the desktop header right away (which might not be optimized) or shouldn't display the desktop header at all (which is a bug)

3 - On both cases, you can see on the "Network" tab, on the `__pickRuntime` requests, that there's a `__device` field on the request, and that only the blocks for the respective devices come through. Also, on the initial `__RUNTIME__` value of the initial HTML request, the same should also apply.

### Depends on:
https://github.com/vtex/pages-graphql/pull/368
https://github.com/vtex/render-server/pull/615
https://github.com/vtex-apps/device-detector/pull/14


react/components/RenderProvider.tsx
```diff
@@ -343,9 +354,15 @@ class RenderProvider extends Component<Props, RenderProviderState> {
   }
 
   public UNSAFE_componentWillReceiveProps(nextProps: Props) {
-    // If RenderProvider is being re-rendered, the global runtime might have changed
-    // so we must update all extensions.
-    if (this.rendered) {
+    const updatedProps = difference(
```
jgfidelis, line 357: Why do a whole diff, and not only compare nextProps.runtime !== this.props.runtime?

It is usually not a good idea to do such a diff on cWRP, but you already know that

react/utils/routes.ts
```diff
@@ -107,17 +108,23 @@ export const fetchServerPage = async ({
...
   fetcher: GlobalFetch['fetch']
+  deviceInfo?: DeviceInfo
 }): Promise<ParsedServerPageResponse> => {
   const parsedUrl = parse(path)
   parsedUrl.search = undefined
   parsedUrl.query = {
     ...rawQuery,
     __pickRuntime: runtimeFields,
+    ...(deviceInfo && {
+      __device: deviceInfo.type,
```
jgfidelis, line 124: is this a new querystring param? If yes, we need to add it to the CDN whitelist

react/components/RenderProvider.tsx
lbebber: i did this because i think it is safer, and cWRP on this component is seldom called, and the diff is not that costly anyway

I'll look into it tho

react/utils/routes.ts
lbebber: you mean this? https://github.com/vtex/render-server/pull/615/files#diff-2ce406e9cb901ca2f74fd823587990c7R5

react/components/RenderProvider.tsx
```diff
@@ -57,6 +57,8 @@ import ExtensionManager from './ExtensionPoint/ExtensionManager'
 import ExtensionPoint from './ExtensionPoint'
 import { RenderContextProvider } from './RenderContext'
 import RenderPage from './RenderPage'
+import { withDevice, WithDeviceProps, DeviceInfo } from '../utils/withDevice'
+import difference from 'ramda/es/difference'
```
jgfidelis, line 61: import `{difference} from 'ramda'`

react/components/RenderProvider.tsx
lbebber: I've measured it, it takes about 1~3ms, and rarely runs.

react/utils/routes.ts
lbebber: (narrator: he didn't mean that)

gonna talk to frederico about it

brunoabreu (approved):

PR #524, opened 2020-05-20
Feat/prefetch

jgfidelis:
#### What does this PR do? \*

Before, as soon as we downloaded the query data for each path, we would hydrate the apollo cache.

This lead to problems, probably bugs in react-apollo, that made the vtex.search (biggy search) to lose its content when doing the prefetch of products inside a search page.

#### How to test it? \*

www.exito.com/search?_query=tv&workspace=pref
https://www.als.com/?workspace=pref

confirm there are prefetch happening and that everything works as expected

### DIFF
--- a/react/utils/flags.ts
+++ b/react/utils/flags.ts
@@ -1,7 +1,7 @@
 const flags = {
   RENDER_NAVIGATION: true,
   VTEX_ASSETS_URL: true,
-  PREFETCH: false,
+  PREFETCH: true,
 }
 
 window.flags = flags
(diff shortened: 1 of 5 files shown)


react/utils/flags.ts
```diff
@@ -1,7 +1,7 @@
 const flags = {
   RENDER_NAVIGATION: true,
   VTEX_ASSETS_URL: true,
-  PREFETCH: false,
+  PREFETCH: true,
```
lbebber, line 4: :developers:

PR #527, opened 2020-05-29
Replace node URL module with window.URL

kaisermann:
#### What does this PR do? \*

The url module used by the routes.ts is a node module. This means that only by just importing it, our builder-hub's webpack adds 1580+- lines of code (or 5kb gzipped) of unneeded code to our stores javascript payload. We can easily replace it with the native window.URL constructor.

Note: IE 11 doesn't have the `URL` and `URLSearchParams` constructors, but we're already polyfilling it [here](https://github.com/vtex/render-server/blob/master/node/utils/assetLoader.ts#L56).

Story: https://app.clubhouse.io/vtex/story/38530/remove-url-module-from-render-runtime

image

#### How to test it? \*

1) Go to https://kiwi--storecomponents.myvtex.com
2) Navigate to your heart's desire

Also tested the IE 11 polyfill via browserstack.

n/a

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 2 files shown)


lbebber:
LGTM but please don't release this on a friday 😂

PR #528, opened 2020-05-29
use own LRUCache

jgfidelis:
#### What does this PR do? \*
The lru-cache lib was meant to be use in a node env, it was breaking on IE11.

Reimplement it with a doubly linked list

#### How to test it? \*

teste--storecomponents.myvtex.com


react/components/Prefetch/LRUCache.ts
```diff
@@ -0,0 +1,124 @@
+class Node<T> {
```
kaisermann, line 1: I hate to be that person... but since we're adding a custom LRU lib, we should at least create some tests for it.

react/components/Prefetch/PQueue.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
```
kaisermann, line 1: why disable eslint?

react/hooks/prefetch.ts
```diff
@@ -255,32 +255,30 @@ export const usePrefetchAttempt = ({
...
-            page,
-            pages,
-            prefetchState,
-            hints,
-            renderMajor,
-            validCache,
-          }),
-        { priority }
-      )
+      // queue.add(
```
kaisermann, line 258: Why is this commented? Should it still be here?

react/components/Prefetch/PrefetchContext.tsx
```diff
@@ -50,17 +48,13 @@ const HALF_HOUR_MS = 1000 * 60 * 30
...
-    search: new LRUCache({ max: 75, dispose: disposeFn, maxAge: HALF_HOUR_MS }),
-    other: new LRUCache({ max: 75, dispose: disposeFn, maxAge: HALF_HOUR_MS }),
+    product: new LRUCache({ max: 100, disposeFn, maxAge: HALF_HOUR_MS }),
+    search: new LRUCache({ max: 75, disposeFn, maxAge: HALF_HOUR_MS }),
+    other: new LRUCache({ max: 75, disposeFn, maxAge: HALF_HOUR_MS }),
   },
   pathsState: {},
   routePromise: {},
-  queue: new PQueue({ concurrency: MAX_CONCURRENCY, autoStart: false }),
+  queue: new PQueue(),
```
kaisermann, line 57: Is this `PQueue` something WIP?

react/components/Prefetch/PQueue.ts
jgfidelis: it was complaining about the empty methods

react/components/Prefetch/PrefetchContext.tsx
jgfidelis: yes, we could not solve, this previous lib we used was not meant to be used in a browser and did not have a valid es5 implementation. Someone that wants to put this in production, will need to solve this on its own.

react/components/Prefetch/LRUCache.ts
jgfidelis: tests added

PR #540, opened 2020-06-24
Feat/prefetch by setting

jgfidelis:
#### What does this PR do? \*

go to
https://storetheme.vtex.com/?workspace=teste2

open network tab, see prefetch does not happen

`vtex settings set vtex.store enablePrefetch true`

reload page, see it happens

### DIFF
--- a/react/utils/routes.ts
+++ b/react/utils/routes.ts
@@ -213,7 +213,7 @@ const promiseWithCounterWrapper = <T = any>(
 
 export const isPrefetchEnabled = (
   storeSettings: Record<string, any> | null | undefined
-) => isEnabled('PREFETCH') && storeSettings?.enablePrefetch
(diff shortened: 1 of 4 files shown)


react/utils/routes.ts
```diff
@@ -211,10 +211,16 @@ const promiseWithCounterWrapper = <T = any>(
     })
 }
 
-export const isPrefetchActive = () =>
+export const isPrefetchEnabled = (
+  storeSettings: Record<string, any> | null | undefined
+) => isEnabled('PREFETCH') && storeSettings?.enablePrefetch
```
brunoabreu, line 216: It is weird when a `isSomething()` function returns `undefined`.

jgfidelis: do you even javascript bruh?

PR #541, opened 2020-06-25
Fix critical

tlgimenes:
#### What does this PR do? \*
The current implementation of critical has the following algorithm
```
1. read style references from __RUNTIME__ global variable
2. for each style, fetch it's content
3. for each fetched content create a <style> tag with the fetched content as innerHtml
```

This implementation has a problem when there is a CSS3 `url()` inside the style's source. 

For example. Suppose we have a `url('my-image.png')` inside a `style.css` sheet that is available under the path `/assets`. 
The aforementioned algorithm will create the following style tag
```
<style>
url('my-image.png')
</style>
```

This style tag will make the browser fetch `my-image.png` under the wrong path `/my-image.png`, instead of  the correct path `/assets/my-image.png`.

To solve this problem, the following algorithm was tested
```
1. read style references from __RUNTIME__ global variable
2. for each style, create a <link rel="preload"> tag
3. for each link tag created, listen to the "onload" and "onerror" events
4. wait for all onloads/onerrors events of the generated link tags to fire.
5. for each link tag, set the rel attribute to "stylesheet"
```

The proposed algorithm solves the `url()` problem by using the browser's mechanism to solve assets path and later "hydrating" all stylesheets at once. However, this approach is still not correct, since we don't necessarily wait for the uncritical styles to be applied before hydrating React, what can lead to unstyled components being rendered on the screen.

The final solution we've come with was to make a hybrid approach using Render Server, such that steps 1,2 and 3 of the previous solutions would be done one the server and hard coded into the html, and render-runtime would:
```
1.  wait for all onloads/onerrors events of the uncritical link tags to fire.
2. just before react hydration, for each link tag, set the rel attribute to "stylesheet"
3. hydrate react
```

This final approach increases the "First Meaningful Paint", however renders react correctly. 

#### How to test it? \*
just open a production workspace and
```
vtex install vtex.render-runtime@8.110.1-beta.7 vtex.render-server@8.127.0-beta.14 vtex.critical-css-extractor@0.5.0-beta.6 && vtex settings set vtex.store enableCriticalCSS true && vtex browse
```

check that the critical tag is present on the home's html


manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "render-runtime",
-  "version": "8.115.0",
+  "version": "8.116.1-beta.0",
```
lbebber, line 4: cleanup beta version before merging

lbebber (approved):
just 1 small thing

PR #542, opened 2020-06-30
Feature/unveilhooks

lbebber:
#### What does this PR do? \*

Upgrades lazysizes plugin and adds unveilhooks plugin, which allows lazyloading background images.
(Takes up 1.4kb in size)

Can be seen working here https://lbebber--worldwidegolf.myvtex.com/

In order to see it in action, hover any item on the header menu, and notice that its images are loaded lazily.

### DIFF
--- a/react/externals.json
+++ b/react/externals.json
@@ -53,7 +53,11 @@
       },
       {
         "clientOnly": true,
-        "path": "npm:lazysizes@5.1.0/lazysizes.min.js"
+        "path": "npm:lazysizes@5.2.2/plugins/unveilhooks/ls.unveilhooks.min.js"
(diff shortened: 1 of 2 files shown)


react/externals.json
```diff
@@ -53,7 +53,11 @@
       },
       {
         "clientOnly": true,
-        "path": "npm:lazysizes@5.1.0/lazysizes.min.js"
+        "path": "npm:lazysizes@5.2.2/plugins/unveilhooks/ls.unveilhooks.min.js"
```
lbebber, line 56: Takes up 1.4kb, but allows lazyloading background images, which might make up for it

PR #551, opened 2020-08-10
Feature/simpler blink fix

lbebber:
#### What does this PR do? \*
Simplifies the logic that prevents blinking images on partial hydration.

The intention is to prevent weird react bugs that may happen.

Instead of reusing the previous already-loaded image elements, it just disables lazyloading from the newly rendered ones.

To test, go to this page (without "Disable cache" turned on), and scroll down to the shelf. The images shouldn't blink
(this is the current behaviour already, this PR just simplifies the logic)

https://storetheme.vtex.com/?workspace=lbebberblink1&v=01

### DIFF
--- a/react/components/Hydration/PreventHydration.tsx
+++ b/react/components/Hydration/PreventHydration.tsx
@@ -37,7 +37,7 @@ async function verifyComponentImplementation(
 
   if (retries > 0) {
     const timeout = 1100 - retries * 100
-    await new Promise(resolve => setTimeout(resolve, timeout))
+    await new Promise((resolve) => setTimeout(resolve, timeout))
     const result = await verifyComponentImplementation(
       runtime,
       treePath,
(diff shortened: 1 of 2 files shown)


react/components/Hydration/PreventHydration.tsx
```diff
@@ -152,6 +106,49 @@ const useAssetLoading = (extensionPointId?: string, shouldLoad?: boolean) => {
...
+  useLayoutEffect(() => {
+    if (!shouldMakeImagesEager) {
+      return
+    }
+
+    const images = alreadyLoadedImages.current.map((image) =>
+      safeQuerySelector<HTMLImageElement>(`img[data-src="${image.src}"]`)
+    )
+
+    // Just a sanity check
```
jgfidelis, line 133: nitpick: unneeded comment, but ok

react/components/Hydration/PreventHydration.tsx
```diff
@@ -152,6 +106,49 @@ const useAssetLoading = (extensionPointId?: string, shouldLoad?: boolean) => {
...
+    if (!images) {
+      return
+    }
+
+    for (const image of images) {
+      if (!image) {
+        return
```
lbebber, line 140:
```suggestion
        continue
```

Writing this down here before I forget

PR #556, opened 2020-09-04
Feature/keep workspace

lbebber:
#### What does this PR do? \*
Keeps workspace during navigation if it's set via querystring.
Currently the workspace, if set this way, is reset during navigation and the page is reloaded.

https://www.carrefour.com.br/?ak-testab=vtex&workspace=lbebberwks1&v=01

### DIFF
--- a/react/components/Link.tsx
+++ b/react/components/Link.tsx
@@ -123,9 +123,9 @@ const Link: React.FunctionComponent<Props> = ({
         !isTelephoneUrl(to) &&
         !isMailToUrl(to)
       ) {
-        return appendWorkspace(rootPath + to, workspace)
+        return rootPath + to
       }
-      return appendWorkspace(to, workspace)
+      return to
     }
     if (page) {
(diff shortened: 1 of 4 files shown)


rafaelbac (approved):
LGTM

react/components/Link.tsx
```diff
@@ -25,6 +25,17 @@ interface Props extends NavigateOptions {
...
+const appendWorkspace = (
+  url: string | undefined,
+  workspace: string | undefined
+) => {
+  if (!url || !workspace || workspace === '') {
+    return url
+  }
+
+  const separator = String(url).includes('?') ? '&' : '?'
+  return url + separator + 'workspace=' + workspace
```
brunoabreu, line 37: What if there is already a workspace query string? 👀

brunoabreu (approved):

react/components/Link.tsx
lbebber: what if indeed 👀

PR #559, opened 2020-09-06
Lazy init runtime JSON

lbebber:
Prevents from adding __RUNTIME__ et al as an inline script, instead adding them inline as a <template>.

See: https://github.com/vtex/render-server/pull/660

### DIFF
--- a/react/start/runtime.ts
+++ b/react/start/runtime.ts
@@ -1,0 +1,34 @@
+export const loadRuntimeJSONs = () => {
+  const scripts = window?.document?.querySelectorAll<HTMLTemplateElement>(
+    'template[data-type="json"]'
+  )
+  if (!scripts || scripts.length === 0) {
+    return Promise.resolve()
+  }
+
+  const promises = Array.from(scripts).map(
+    (script) =>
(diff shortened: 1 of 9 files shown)


react/start/runtime.ts
```diff
@@ -0,0 +1,45 @@
+const getValue = (element: HTMLTemplateElement) => {
+  if (typeof element.content === 'undefined') {
+    return element.innerHTML
+  }
+
+  let value = ''
+  const childNodes = element.content.childNodes
```
lbebber, line 7:
```suggestion
  // Using nodeValue instead of innerHTML because it's faster
  // https://stackoverflow.com/questions/21311299/nodevalue-vs-innerhtml-and-textcontent-how-to-choose/21311670#21311670
  // The reason for the loop is because large texts are broken into multiple childNodes
  const childNodes = element.content.childNodes
```

lbebber (commented):
LGTM, só colocar esse comentario pra nao mudarem no futuro pra algo mais simples sem saber o motivo

brunoabreu (approved):

PR #567, opened 2020-09-22
Loading when redirecting

vitorflg:
#### What does this PR do? \*

Add the loading bar on top of the document when redirecting to checkout.

#### How to test it? \*

In the workspace, try to add to cart or use the `go to checkout` button of the minicart. The redirect should occur with a presence of a loading bar on top of the document.

https://vitorflg--storecomponents.myvtex.com/

X

X

### DIFF
--- a/react/components/RenderProvider.tsx
+++ b/react/components/RenderProvider.tsx
@@ -498,6 +498,10 @@ export class RenderProvider extends Component<
     return pageGoBack(history)
   }
 
+  public setPreview = (previewValue: boolean) => {
+    this.setState({ preview: previewValue })
+  }
+
   public setQuery = (
     query: Record<string, any> = {},
     {
@@ -544,14 +548,19 @@ export class RenderProvider extends Component<
       runtime: { rootPath },
     } = this.props
     const { pages } = this.state
-    options.rootPath = rootPath
-    options.modifiers = this.navigationRouteModifiers
 
     this.navigationModifierOptions = {
       ...this.navigationModifierOptions,
       ...options.modifiersOptions,
     }
-    options.modifiersOptions = this.navigationModifierOptions
+
+    options = {
+      ...options,
+      rootPath,
+      modifiers: this.navigationRouteModifiers,
+      modifiersOptions: this.navigationModifierOptions,
+      setPreview: this.setPreview,
+    }
 
     if (this.navigationState.isNavigating) {
       const lastOptions = this.navigationState.lastOptions!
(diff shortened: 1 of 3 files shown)


react/utils/pages.ts
```diff
@@ -325,18 +325,29 @@ export function navigate(
...
+
     return true
   }
 
   if (navigationRoute == null) {
     return false
   }
 
+  if (fallbackToWindowLocation) {
+    setPreview && setPreview(true)
```
kaisermann, line 344:
```suggestion
    if(setPreview) setPreview(true)
```

react/components/RenderProvider.tsx
```diff
@@ -544,14 +548,19 @@ export class RenderProvider extends Component<
...
-    options.modifiers = this.navigationRouteModifiers
 
     this.navigationModifierOptions = {
       ...this.navigationModifierOptions,
       ...options.modifiersOptions,
     }
-    options.modifiersOptions = this.navigationModifierOptions
+
+    options = {
+      ...options,
```
klzns, line 558: I think it's better to keep how it was, without destructuring. I'm afraid of the side effects this might cause, performance-wise and behaviour-wise

klzns (changes requested):

kaisermann (changes requested):
One tiny nitpick

and what is `setPreview` the best name for this behavior? I'm not familiar with the `preview` terminology used in the project, but would something like `showPageLoading` be a better fit?

react/utils/pages.ts
victorhmp, line 344: How about

```suggestion
   setPreview?.(true)
```

victorhmp (changes requested):

react/utils/pages.ts
vitorflg: hmmm, good point. Do you recommend creating 2 functions, showPageLoading and hidePageLoading or were you thinking to keep the param (true/false) to set the preview state?

vitorflg: I liked the explicit way that kaisermann recommended above, what do you think?

iaronaraujo (approved):

victorhmp (approved):
LGTM 😄

react/utils/pages.ts
victorhmp: That's fine! 😄

icazevedo (approved):

kaisermann (approved):

react/utils/pages.ts
```diff
@@ -320,23 +321,34 @@ export function navigate(
 ) {
   const {
     scrollOptions,
-    fallbackToWindowLocation = true,
+    fallbackToWindowLocation = false,
```
klzns, line 324: Why the default was changed?

klzns (approved):

PR #568, opened 2020-09-23
Better fold + lazy footer

lbebber:
#### What does this PR do? \*
Improves fold implementation and functionality, and adds support for lazy footer rendering, enabled via `enableLazyFooter` setting.

This PR also refactors the `LayoutContainer` class into a Function Component, and also simplifies its implementation a bit.

Before, if the user scrolled reasonably quickly, they would stumble upon the footer, which would promptly blink away, and then the whole rest of the page would render
before-als
before-footer

With this PR, each concealed row adds its own empty spacer div, which brings the scroll area closer to the actual value. It also adds spacing between the above-the-fold content and the footer, which prevents the user from seeing it in most cases.

after-als-quick
after-quick

Also, since it renders the rows one by one, instead of the whole page in a single blow, the rendering should be much smoother, which is more noticeable if the user scrolls more slowly.
after-als-smooth
after-smooth

This PR also adds lazy rendering of the footer, if `enableLazyFooter`  is enabled (via `vtex settings set vtex.store enableLazyFooter true`)

In the example below, JS is disabled. Notice via the scrollbar that the user is scrolling to the bottom of the page, with no footer in sight. This should improve the initial rendering performance.
carrefour-empty

After and before, throttled 4x
Screen Shot 2020-09-23 at 16 32 31

#### How to test it? \*

Can be tested in the following workspaces:
https://www.als.com/?workspace=lbebberfold1&v=01
https://www.exito.com/?workspace=lbebberfold1
https://www.carrefour.com.br/?ak-testab=vtex&workspace=lbebberfold1&v=01

Just refresh and scroll around a few times, with varying speed.

Compare with their respective master workspaces:
https://www.als.com/
https://www.exito.com/
https://www.carrefour.com.br/?ak-testab=vtex

### DIFF
--- a/react/components/LazyRender.tsx
+++ b/react/components/LazyRender.tsx
@@ -9,7 +9,7 @@ interface Props {
 
 const LazyRender: FunctionComponent<Props> = ({
   children,
-  height = 300,
+  height = 400,
   offset = 300,
   debug = false,
 }) => {
--- a/react/components/LazyRender.tsx
+++ b/react/components/LazyRender.tsx
@@ -11,7 +11,6 @@ const LazyRender: FunctionComponent<Props> = ({
   children,
   height = 400,
   offset = 300,
-  debug = false,
 }) => {
   const ref = useRef(null)
   const [hasBeenViewed, setHasBeenViewed] = useState(false)
@@ -20,9 +19,6 @@ const LazyRender: FunctionComponent<Props> = ({
     ref,
     onView: () => {
       setHasBeenViewed(true)
-      if (debug) {
-        console.log('ViewDetector has been viewed')
-      }
     },
     once: true,
     initializeOnInteraction: true,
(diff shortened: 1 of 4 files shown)


brunoabreu (approved):
TYP

react/components/LazyRender.tsx
```diff
@@ -0,0 +1,60 @@
...
+  debug = false,
+}) => {
+  const ref = useRef(null)
+  const [hasBeenViewed, setHasBeenViewed] = useState(false)
+
+  useOnView({
+    ref,
+    onView: () => {
+      setHasBeenViewed(true)
+      if (debug) {
```
kaisermann, line 23: Are we going to leave this log here? Could we add a `process.env.NODE_ENV` check at least?

kaisermann (approved):

PR #574, opened 2020-10-16
Remount lazy loaded images when src changes

kaisermann:
#### What does this PR do? \*

Currently, if a lazyloaded `img` tag has its `src` attribute changed, our monkey patched `img`s are not changing the `src` attribute, only the `data-src`. This PR adds a `key` based on the passed `src` so `React` is re-renders the whole tag when the src changes. 

#### How to test it? \*

**Compare:**
https://kiwi--celsia.myvtex.com/
https://celsia.myvtex.com/

**Flow:**
1) Focus on search bar
2) Type `samsung`
3) Change the sku variation and see if the image changes

image

n/a


klzns (changes requested):
missing changelog

klzns (approved):

PR #584, opened 2020-11-21
Feature/inner fold

lbebber:
#### What does this PR do? \*
Adds support for `__fold__` blocks inside other blocks.

#### How to test it? \*
https://www.carrefour.com.br/notebook-samsung-intel-core-i7-9750h-16gb-1tb-256gb-15-6-windows-10-home-odyssey-np850xbd-xg2br-5923425/p?workspace=lbebberinnerfold1

If you scroll down very quickly, you should see the lazy rendered blocks appearing as they cross into the viewport. This is to confirm that the feature is working properly.
carrefour-innerfould-scrollquick

Compare with the version currently in production, where the content is already rendered by the time you would scroll down (which is nicer lol, but heavier)
carrefour-innerfold-off

If you scroll down a little slower there shouldn't be much shifting around though.
carrefour-innerfold-scrollsmooth

### DIFF
--- a/react/components/ExtensionPoint/index.tsx
+++ b/react/components/ExtensionPoint/index.tsx
@@ -12,6 +12,7 @@ import GenericPreview from '../Preview/GenericPreview'
 import LoadingBar from '../LoadingBar'
 import { LazyImages } from '../LazyImages'
 import LazyRender from '../LazyRender'
+import FoldableContainer from '../FoldableContainer'
 
 // TODO: Export components separately on @vtex/blocks-inspector, so this import can be simplified
 const InspectBlockWrapper = React.lazy(
@@ -65,7 +66,15 @@ export function getChildExtensions(runtime: RenderContext, treePath: string) {
     return isChild && isNotSlot
   })
 
-  return childBlocks.map((child, i) => {
+  /** Rudimentary detection of __fold__ blocks. Shouldn't be a problem
+   * because other types of fold blocks aren't used inside other blocks
+   * anyway. Could be improved if necessary.
+   */
+  const foldIndex = childBlocks.findIndex((child) =>
+    child.extensionPointId.includes('_fold_')
+  )
+
+  const childExtensions = childBlocks.map((child, i) => {
     const childTreePath = mountTreePath(child.extensionPointId, treePath)
 
     const childExtension = runtime?.extensions[childTreePath]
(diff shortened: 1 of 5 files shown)


react/components/FoldableContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+import LazyRender from './LazyRender'
+
+interface Props {
+  foldIndex?: number
+}
+const FoldableContainer: FunctionComponent<Props> = ({
+  children,
+  foldIndex,
+}) => {
+  if (foldIndex == null || foldIndex === -1) {
```
klzns, line 11:
```suggestion
  if (foldIndex == null || foldIndex < 1) {
```

react/components/ExtensionPoint/index.tsx
```diff
@@ -80,6 +89,15 @@ export function getChildExtensions(runtime: RenderContext, treePath: string) {
       />
     )
   })
+
+  if (foldIndex > -1) {
```
klzns, line 93:
```suggestion
  if (foldIndex > 0) {
```

react/components/FoldableContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return <>{children}</>
+  }
+  const childrenArray = React.Children.toArray(children)
+  const aboveTheFold = childrenArray.slice(0, foldIndex)
```
klzns, line 15: Suggesting all of this because if index is zero the above the fold would be an empty array.

klzns (changes requested):

react/components/FoldableContainer.tsx
lbebber: not sure if I get it--i mean, if you put the fold at the very top of the page, where its index would be `0`, thats the expected behaviour I think: the above-the-fold array would be empty, and all the elements would be contained in the below-the-fold array.

right? not sure if i'm missing something

klzns: Hmm, ok, I didn't think about that use case (it's kind of weird though)

klzns (approved):

react/components/FoldableContainer.tsx
lbebber: it is probably not a very good use case indeed haha but it would be the expected behaviour of doing so

PR #595, opened 2021-01-11
Extend prefetch scenarios

vitorflg:
#### What does this PR do? *

The prefetch queue listener is not triggered when the document has already been loaded. So, in this case, instead of adding a listener, we could just start the queue to make prefetch work in all scenarios.

#### How to test it? \*

https://vitorflg--storecomponents.myvtex.com/

Prefetch should work in all scenarios.

#### Describe alternatives you've considered if any. \*

X

X

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


brunoabreu (approved):

icazevedo (approved):

klzns (approved):
Nice, do not merge today though

PR #597, opened 2021-01-19
Disable folding on the site editor

vitorflg:
#### What does this PR do? \*

Blocks that are under fold block sometimes not render in the blocks list of the site editor. It disables that performance feature in this scenario.

#### How to test it? \*

https://vitorflg--muji.myvtex.com/admin/cms/site-editor

Create a different component/logic flow for the site editor and the end-client render, but it would need a big refactoring.

X


lbebber (approved):
LGTM

PR #600, opened 2021-01-22
Disable hydrate-on-view on the site editor

vitorflg:
#### What does this PR do? \*

Change the site editor identifier. A flag added to the admin-pages now permits us to better recognize if a page is being requested in the admin or not.

#### How to test it? \*

https://vitorflg--muji.myvtex.com/admin/cms/site-editor

X

X

### DIFF
--- a/react/components/ExtensionPoint/ComponentLoader.tsx
+++ b/react/components/ExtensionPoint/ComponentLoader.tsx
@@ -129,8 +129,10 @@ const ComponentLoader: FunctionComponent<Props> = (props) => {
     <AsyncComponent {...asyncComponentProps}>{children}</AsyncComponent>
   )
 
-  const shouldHydrate =
-    hydration ||
+  const isSiteEditor = !runtime?.route?.path.includes('__siteEditor')
+
+  const shouldLazyHydrate =
+    !hydration ||
     hydration === 'always' ||
     /** TODO: Currently it only applies partial hydration on top level components
      * Nested partial hydration should be supported in the future */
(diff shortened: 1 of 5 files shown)


react/components/ExtensionPoint/ComponentLoader.tsx
```diff
@@ -129,15 +130,15 @@ const ComponentLoader: FunctionComponent<Props> = (props) => {
   )
 
   const shouldHydrate =
-    !hydration ||
+    hydration ||
```
brunoabreu, line 133: It seems weird.. 

`hydration || hydration === 'always'`.. any value of `hydration` will work here.

vitorflg: It is weird! I will see with @lbebber if can just do `let shouldHydrate = hydration || ..`

react/components/ExtensionPoint/ComponentLoader.tsx
```diff
@@ -128,7 +129,9 @@ const ComponentLoader: FunctionComponent<Props> = (props) => {
     <AsyncComponent {...asyncComponentProps}>{children}</AsyncComponent>
   )
 
-  const shouldHydrate =
+  const isSiteEditor = !runtime?.route?.path.includes('__siteEditor')
+
+  const shouldLazyHydrate =
```
lbebber, line 134: for this to be renamed, the logic would need to be inverted (if this variable is true, it will hydrate eagerly, not lazily)

lbebber (changes requested):

brunoabreu (approved):

lbebber (approved):
#agorafoi

PR #608, opened 2021-04-19
Decoding parameters to avoid double encoding while using route-parser

FabianaFerreira:
#### What does this PR do? \*
It fixes a bug in basically every production store when searching three terms. The requests were being blocked by the WAF that we have on Azion, because they contained double encoded strings. You can simulate the behavior, for example, going to polishop.com.br and trying to search for "fritadeira air fryer". Once you go to the Network tab, you'll see that you receive a `400`, as the image below:
image

You can check in the following image that, instead of having `fritadeira%20air%20fryer` in the path, you will find `fritadeira%2520air%2520fryer`. This is due to the fact that the package that we use to join the path using the terms object already encodes the string. However, the strings were already reaching the RouteParser's method encoded.

image

In order for us to fix the problem without removing the usage of the `route-parser` package, we decided to implement a util function that decodes the string before using it in the method. 

> It's important to emphasize that the behavior of blocking the requests only affects searches with three terms or more because it's based on a score and if it goes above the threshold, it will block the request

#### How to test it? \*
In the Store Theme public domain ([storetheme.vtex.com](storetheme.vtex.com)), you can check that the problem is "solved". In this account, there is a beta version of render-runtime installed. If you install the most recent stable version (`vtex.render-runtime@8.128.2`), you'll be able to reproduce the problem that was mentioned below.

I've also created some unit tests in order to reproduce the double encoding and to test the function that I've implemented.

As a temporary solution to the client that reported the problem, we deactivated the WAF, but this is not the best scenario, because it disables every protection that the client's public domain has.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Double encoding while generating paths
 
 ## [8.128.2] - 2021-03-18
 
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "render-runtime",
-  "version": "8.128.2",
+  "version": "8.128.3-beta.0",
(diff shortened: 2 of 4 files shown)


klzns (changes requested):
Missing changelog

klzns (approved):

PR #613, opened 2021-06-11
Adding internal params identifiable via __ prefix

icazevedo:
#### What does this PR do? \*
This PR adds a filter to Link params. This filter prevents all params that start with `__` from being considered during path transformations. This was added to safely allow the usage of params that are sent between pages but shouldn't be added to the query params due to user experience compromises.

#### How to test it? \*
[Workspace](http://icarorender--storecomponents.myvtex.com)
[Workspace 2](http://icarogtm--storecomponents.myvtex.com)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -621,7 +621,7 @@ The `Link` React component is responsible for rendering an `a` HTML element that
 | :------------- |:-------------| :-----|:-----|
 | `page`     | `string`  | The name of the page that the user will be redirected to. Maps to a `blocks.json` block (e.g., `'store.product'`)||
 | `to`     | `string`    |The URL of the page that the user will be redirected to (e.g., `/shirt/p?skuId=1`). Notice that `to` is an **alternative** to `page` and it contains the whole URL instead of the page name. | |
-| `params` | `object`      | The `param` values of the page path in a key-value format (e.g, `{slug: 'shirt'}`). | `{}`|
+| `params` | `object`      | The `param` values of the page path in a key-value format (e.g, `{slug: 'shirt'}`). Params that starts with `__` are not considered on path transformations, and can be generaly be used as an alternative to query params | `{}`|
 | `query` | `string`  | The representation of the query params that are appended to the page path (e.g., `skuId=231`.) | `''` |
 | `onClick` | `function` | A callback that is fired when the user clicks on a component (e.g., `() => alert('Salut')`) | |
 | `replace` | `boolean` | The boolean value used to indicate if it should call (`true`) the replace function to navigate or not (`false`) | |
--- a/docs/README.md
+++ b/docs/README.md
@@ -628,7 +628,7 @@ The `Link` React component is responsible for rendering an `a` HTML element that
 
 Other props you pass will be forwarded to the `a` component and can be used for customization.
 
-Take the following usage example:
+Take the following usage examples:
 
 ```tsx
 import React from 'react'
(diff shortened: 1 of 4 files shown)


react/components/RenderProvider.tsx
```diff
@@ -654,6 +654,19 @@ export class RenderProvider extends Component<
     }
   }
 
+  private mergeRouteParams<T extends { params: any }>(
+    matchingPage: T,
+    transientRoute: { params: any }
```
igorbrasileiro, line 659: Why this property `transientRoute` isn't of type `T`?

react/components/RenderProvider.tsx
```diff
@@ -654,6 +654,19 @@ export class RenderProvider extends Component<
...
+  ): T {
```
igorbrasileiro, line 660: We can be more accurate in typing the return saying that params are an object?

react/components/RenderProvider.tsx
```diff
@@ -654,6 +654,19 @@ export class RenderProvider extends Component<
...
+    return {
+      ...matchingPage,
+      params: {
+        ...transientRoute.params,
```
igorbrasileiro, line 664: If you can use spread, transientRoute is an object, probably T should extend something like that

```ts
T extends { params: Record<string | number | symbol, any> }
```

docs/README.md
```diff
@@ -621,7 +621,7 @@ The `Link` React component is responsible for rendering an `a` HTML element that
 | :------------- |:-------------| :-----|:-----|
 | `page`     | `string`  | The name of the page that the user will be redirected to. Maps to a `blocks.json` block (e.g., `'store.product'`)||
 | `to`     | `string`    |The URL of the page that the user will be redirected to (e.g., `/shirt/p?skuId=1`). Notice that `to` is an **alternative** to `page` and it contains the whole URL instead of the page name. | |
-| `params` | `object`      | The `param` values of the page path in a key-value format (e.g, `{slug: 'shirt'}`). | `{}`|
+| `params` | `object`      | The `param` values of the page path in a key-value format (e.g, `{slug: 'shirt'}`). Params that starts with `__` are not considered on path transformations, and can be generaly be used as an alternative to query params | `{}`|
```
igorbrasileiro, line 624: Can you show me an example of this change? I have no context about it.

igorbrasileiro (approved):
Left some comments, but lgtm

react/components/RenderProvider.tsx
icazevedo: Because typing it as T makes the function expect that `matchingPage` and `transientRoute` have the same type, which doesn't always apply.

react/components/RenderProvider.tsx
icazevedo: I thought about it but this is how params it's typed on the rest of the code, so 🤷

react/components/RenderProvider.tsx
icazevedo: I thought about it but this is how params it's typed on the rest of the code, so 🤷

docs/README.md
icazevedo: The `__listName` property I'm sending from Product Summary is an example of that. Do you think I should add or reference it to the docs?

victorhmp: I think so! An example is always great 😄

victorhmp (approved):

PR #614, opened 2021-06-22
Hydration flag

lbebber:
#### What does this PR do? \*

Adds flag that signals that hydration is done. Intended for using with lazy pixel loading

Refer to https://github.com/vtex/render-server/pull/723/ for more details

### DIFF
--- a/react/core/main.tsx
+++ b/react/core/main.tsx
@@ -379,9 +379,13 @@ function start() {
       }))
     } else {
       setLazyCookie(runtime.workspaceCookie)
-      maybeRenderPromise.then(() => {
-        ;(window as any).__HAS_HYDRATED__ = true
-      })
+      maybeRenderPromise
+        .then(() => {
+          ;(window as any).__HAS_HYDRATED__ = true
+        })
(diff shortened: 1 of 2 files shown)


react/core/main.tsx
```diff
@@ -379,6 +379,9 @@ function start() {
       }))
     } else {
       setLazyCookie(runtime.workspaceCookie)
+      maybeRenderPromise.then(() => {
```
brunoabreu, line 382: What if it throws some error? Is this possible? 👀

brunoabreu (approved):

react/core/main.tsx
lbebber: hmmmm :thonking:

should not be possible, and I don't think anything bad would happen here in that case, but I'll investigate

lbebber: Screen Shot 2021-06-23 at 17 34 22
All it does is complain about an uncaught error in a promise. 

That said, if this promise fails, there are bigger problems--hydration has failed.

_That_ said, I will treat this error anyway

PR #625, opened 2021-09-24
Feature/increase perf measure samples

lbebber:
#### What does this PR do? \*
Increases the default log sampling rate for the performance measures.

Also allows highlighting certain accounts and pages to increase the sample rate for them specifically, for closer study.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Changed
-- Increase the performance measurement sampling to 1%.
+- Increase the default performance measurement sampling to 1%, and for specific accounts and pages to 4%.
 
 ## [8.132.0] - 2021-09-23
 ### Added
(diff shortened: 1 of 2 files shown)


react/components/withPerformanceMeasures.tsx
```diff
@@ -0,0 +1,247 @@
...
+  )
+  if (shouldDebugLogMeasures) {
+    return true
+  }
+
+  const DEFAULT_LOG_SAMPLING_RATE = 0.0075
+  // Allows increasing the log rate of certain accounts and pages for closer analysis.
+  const HIGHLIGHT_LOG_SAMPLING_RATE = 0.04
+  const HIGHLIGHT_ACCOUNTS = ['carrefourbr']
+  const HIGHLIGHT_PAGES = ['store.home']
```
lbebber, line 132: The change is obscured because the diff got messed up due to moving this code to a different file, but this is one of the main changes:

Increases sampling rate in general, and for the carrefourbr home in particular

react/components/withPerformanceMeasures.tsx
```diff
@@ -0,0 +1,247 @@
...
+    useEffect(() => {
+      if (!hasHydrated) {
+        setHasHydrated(true)
+      }
+    }, [setHasHydrated, hasHydrated])
+
+    if (hasHydrated) {
+      // Using setTimeout to force the execution of this function slightly after
+      // the entire first post-render task (i.e. after all the useEffects etc).
+      setTimeout(() => {
```
lbebber, line 224: Also this is an important, if a bit esoteric, change.

PR #637, opened 2022-05-25
Update react to version 17.0.2

matheusps:
#### What does this PR do?

- Update react version to 17.0.2

#### How to test it?

- Workspace: https://react17--recorrenciaqa.myvtex.com/.

#### Describe alternatives you've considered if any.

- Adding it as an "alternative".

#### Unresolved questions

- Are there any other modules that I should be concerned about?
- Could this impact users in an unexpected way?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Update react version to `17.0.2`.
+
 ## [8.132.4] - 2022-03-17
 ### Changed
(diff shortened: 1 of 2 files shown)


emersonlaurentino (approved):

kevinch (approved):
LGTM but i'm not a render-runtime expert.

lbebber (approved):
I've run through the [list of potential breaking changes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes) and nothing rings a bell.
it's impossible to guarantee, but it seems as safe as it gets

just keep an eye out for a couple of days after deploy to see if nothing weird comes up that might be related to this change!

PR #642, opened 2022-11-16
fix(NoSSR): add children typings

EduardoRodriguesF:
#### What does this PR do? \*

The NoSSR component was missing `children` from its typings, forcing custom components to ts-ignore it.
Children prop already works, this PR simply adds it to Props interface.

#### How to test it? \*

Import NoSSR from a typescript file and pass any children to it. If everything is fine, TypeScript should not complain anymore.

Not sure if this fits the question, but I set it as an optional prop, not sure if it should be mandatory.

Typings

### DIFF
--- a/react/components/NoSSR.tsx
+++ b/react/components/NoSSR.tsx
@@ -20,6 +20,7 @@ const useSSR = () => {
 
 interface Props {
   onSSR?: React.ReactNode
+  children?: React.ReactNode
 }
 
 const NoSSR: FunctionComponent<Props> = ({ children, onSSR }) => {


emersonlaurentino:
This is redundant. `children` is not necessary as the Props interface is called on `FunctionComponent` which has children as a default parameter.

emersonlaurentino:
@EduardoRodriguesF I don't know which version you are using, but I just cloned the project in the latest version and the FunctionComponent is different from its print, here's how it is:

If you look at the `PropsWithChildren` which is responsible for ensuring that there is a type for the `children` prop.

Screenshot 2023-06-13 at 15 46 39

Also, I made sure that it actually works as shown in the screenshot below:

Screenshot 2023-06-13 at 15 50 17

EduardoRodriguesF:
Interesting... I verified @types/react on my app and it was using 18.x, while this one uses 16.x.

After downgrading, everything works fine! Thanks for your time, @emersonlaurentino :)

On this React version topic, I see that the ideal version to work with VTEX IO is 16.x. Is there any plans for an upgrade on that part for newer versions?

PR #656, opened 2024-07-18
Feat: DIV IDs to elements

arturmagalhaesjr:
#### What does this PR do? \*

Added an ID to DIV elements to map in the CSS elements not rendered by FOLD functionality, so it's possible to create CSS to prevent potential CLS.

**This ID is based on the block name from the theme.**

#### How to test it? \*

https://artcls--hunterdouglasnl.myvtex.com/houten-jaloezieen

image


react/components/LayoutContainer.tsx
```diff
@@ -43,8 +44,11 @@ const Container: FunctionComponent<ContainerProps> = ({
     if (elements === '__children__') {
       return <>{children}</>
     }
+    const id = Buffer.from(elements, 'binary').toString('base64')
+    const classNameWithId = `${isRow ? '' : className + ' '}id-${id}`
```
iago1501, line 48: Hi @arturmagalhaesjr , I would still go for the css handles solution, there's no problem to add the lib for the first time there and increase the PR size, the most important is to keep our standarts/patterns in new classes, so, I would go for this implementation:

https://developers.vtex.com/docs/apps/vtex.css-handles

The first example of usage with useCssHandles is already enough for that, something like:

const CSS_HANDLES = ['container'] as const
const { handles } = useCssHandles(CSS_HANDLES, { classes })

...
<div className={handles.container}>

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Added a ID to DIV elements to map in the css elements not rendered by FOLD functionality, so it's possible to create CSS to prevent potential CLS
```
iago1501, line 12:
```suggestion
- Added a class to `<div>` elements to map in the CSS elements not rendered by __FOLD__ functionality, so it's possible to create CSS to prevent potential CLS
```
```suggestion
- Added a class to `<div>` elements to map in the CSS elements not rendered by `__FOLD__` functionality, so it's possible to create CSS to prevent potential CLS
```

iago1501 (changes requested):

react/components/LayoutContainer.tsx
iago1501: Yes, there's no problem with adding this new dependency, the main point here is to keep our code standards if necessary, and in this case, we should follow our patterns since it's necessary.

It's better to do it now than in the future if we want to add a new CSS class anywhere else since this part of the code will be forgotten

iago1501: @arturmagalhaesjr , did you have time to review it?

iago1501:
I'm closing this PR since the changes were requested a month ago, so, if this change is still necessary I recommend opening this PR again, add the change, and open a new PR prioritization flow

iago1501:
@arturmagalhaesjr , just checked here, the css handles are not really going to work, so, can you give me an example of this functionality working again? Your workspace is crashed

PR #658, opened 2024-08-08
chore: add admin information to the sentry capture exception event

lucasaarcoverde:
#### What does this PR do? \*

Add additional information to the sentry admin instrumentation, like: `account`, `workspace`, `device`, whether is in `production`  or not, `locale`, and `path`. 

#### How to test it? \*

In [this workspace](https://lucasan--storecomponents.myvtex.com/admin/catalog) you can open the `console.log` and check the data sent whenever an error on production occurs.

Screenshot 2024-08-08 at 10 05 04

### DIFF
--- a/react/error.tsx
+++ b/react/error.tsx
@@ -1,7 +1,7 @@
 /* global module */
 import React, { Component, Fragment } from 'react'
 import ReactJson from 'react-json-view'
-import { captureException } from '@sentry/react'
+import { captureException, getCurrentScope } from '@sentry/react'
 
 require('myvtex-sse')
 
@@ -17,25 +17,51 @@ class ErrorPage extends Component {
   public state = { enabled: false }
 
   public componentDidMount() {
-    window.setTimeout(() => {
-      this.setState({ enabled: true })
-    }, 5000)
+    try {
+      const error = window?.__ERROR__
+      const requestId = window?.__REQUEST_ID__
+      const defaultError =
+        'Render Runtime renderered an error page and there is no error or request id available'
 
-    if (!isAdmin()) {
-      return
-    }
+      const runtime = window?.global?.__RUNTIME__ ?? {}
 
-    const error = window?.__ERROR__
-    const requestId = window?.__REQUEST_ID__
-    const defaultError =
-      'Render Runtime renderered an error page and there is no error or request id available'
+      const {
+        account = 'N/A',
+        workspace = 'N/A',
+        culture: { locale } = { locale: 'N/A' },
+        route: { path } = { path: 'N/A' },
+        loadedDevices = ['N/A'],
+        production = 'N/A',
+      } = runtime
 
-    if (error) {
-      captureException(error)
-    } else if (requestId) {
-      captureException(requestId)
-    } else {
-      captureException(defaultError)
+      const errorInfo = {
+        account,
+        workspace,
+        locale,
+        path,
+        device: loadedDevices[0],
+        production,
+      }
+
+      getCurrentScope().setExtras(errorInfo)
+
+      window.setTimeout(() => {
+        this.setState({ enabled: true })
+      }, 5000)
+
+      if (!isAdmin() || production === false) {
+        return
+      }
+
+      if (error) {
+        captureException(error)
+      } else if (requestId) {
+        captureException(requestId)
+      } else {
+        captureException(defaultError)
+      }
+    } catch (e) {
+      captureException(e)
     }
   }
 
--- a/react/error.tsx
+++ b/react/error.tsx
@@ -17,6 +17,10 @@ class ErrorPage extends Component {
   public state = { enabled: false }
 
   public componentDidMount() {
+    window.setTimeout(() => {
+      this.setState({ enabled: true })
+    }, 5000)
+
     try {
       const error = window?.__ERROR__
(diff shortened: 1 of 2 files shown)


react/error.tsx
```diff
@@ -17,25 +17,51 @@ class ErrorPage extends Component {
...
+
+      const runtime = window?.global?.__RUNTIME__ ?? {}
+
+      const {
+        account = 'N/A',
+        workspace = 'N/A',
+        culture: { locale } = { locale: 'N/A' },
+        route: { path } = { path: 'N/A' },
+        loadedDevices = ['N/A'],
+        production = 'N/A',
```
marcelovicentegc, line 34: _["The null value represents the intentional absence of any object value."](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null)_ :hugs: 

```suggestion
        account = 'null,
        workspace = null,
        culture: { locale } = { locale: null },
        route: { path } = { path: null },
        loadedDevices = [null],
        production = null,
```

marcelovicentegc: Just a suggestion! `"N/A"` LGTM

react/error.tsx
```diff
@@ -17,25 +17,51 @@ class ErrorPage extends Component {
...
+        workspace,
+        locale,
+        path,
+        device: loadedDevices[0],
+        production,
+      }
 
-    if (!isAdmin()) {
-      return
-    }
```
marcelovicentegc, line 26: I would move this to the top, right before the:

```js
window.setTimeout(() => {
      this.setState({ enabled: true })
    }, 5000)
```

marcelovicentegc (changes requested):
@lucasaarcoverde we should leave the `try catch` execution and any logging logic available only on Admin environments. 

The way it's implemented now, we would prepare an error object to send to Sentry, regardless if it's under an Admin environment. Also, we would end up logging errors from the Storefront if the try catch fails for any reason.

react/error.tsx
lucasaarcoverde: good catch! I agree

vmourac-vtex: I would  also suggest extracting the creation of `errorInfo` to private method. This is purely to decrease the size of the method and maintain readability.
Code LGTM, just a suggestion, too

```ts
  private extractErrorInfo(runtime: any): any {
    const {
      account = null,
      workspace = null,
      culture: { locale } = { locale: null },
      route: { path } = { path: null },
      loadedDevices = [null],
      production = null,
    } = runtime

    const errorInfo = {
      account,
      workspace,
      locale,
      path,
      device: loadedDevices[0],
      production,
    }

    return errorInfo
  }
  ```
  
  And call it when we add the extras to the current scope
   ```ts
      const errorInfo = this.extractErrorInfo(runtime);
      getCurrentScope().setExtras(errorInfo)
   ```

**Comment by vmourac-vtex** [comment]
After suggestions from @marcelovicentegc I should be perfect. I'll leave the approval as my comment was purely a suggestion

**Review by vmourac-vtex — APPROVED** [review]

File: react/error.tsx (hunk 2 above)
Comment (lucasaarcoverde) [reply react/error.tsx:26]: Do you mean something like that?

```ts
    window.setTimeout(() => {
      this.setState({ enabled: true })
    }, 5000)
   
    try {
      const error = window?.__ERROR__
      const requestId = window?.__REQUEST_ID__
      const defaultError =
        'Render Runtime renderered an error page and there is no error or request id available'

      const runtime = window?.global?.__RUNTIME__ ?? {}
      const errorInfo = this.extractErrorInfo(runtime)

      if (!isAdmin() || errorInfo.production === false) {
        return
      }

      getCurrentScope().setExtras(errorInfo)

      if (error) {
        captureException(error)
      } else if (requestId) {
        captureException(requestId)
      } else {
        captureException(defaultError)
      }
    } catch (e) {
      if (!isAdmin()) return

      captureException(e)
    }
```

react/error.tsx
```diff
@@ -21,22 +21,53 @@ class ErrorPage extends Component {
...
+      production = null,
+    } = runtime
+
+    const errorInfo = {
+      admin_account: account,
+      admin_workspace: workspace,
+      admin_locale: locale,
+      admin_path: path,
+      admin_device: loadedDevices[0],
+      admin_production: production,
```
marcelovicentegc, line 67: It occurs to me that we are better off with regular `account`, `workspace`, etc nomenclature as it's the one used across other o11y services

```suggestion
      account,
      workspace,
      locale,
      path,
      device: loadedDevices[0],
      production,
```

lucasaarcoverde: I've changed that because of Sentry's documentation
> Some tags are automatically set by Sentry. We strongly recommend against overwriting these [tags](https://docs.sentry.io/concepts/search/searchable-properties/#search-properties). Instead, name your tags with your organization's nomenclature. If you overwrite an automatically set tag, you must use [explicit tag syntax](https://docs.sentry.io/concepts/search/#explicit-tag-syntax) to search for it.

Sentry already has `device` and `production` tags for example.
[Sentry tags documentation](https://docs.sentry.io/platforms/javascript/enriching-events/tags/)

marcelovicentegc: Nice, good catch!!!

marcelovicentegc (approved):

PR #667, opened 2024-09-11
feat: improve UX for error pages

anitavincent:
#### What does this PR do? \*

Implements new UX for error pages.

#### How to test it? \*

Visit https://anita--storecomponents.myvtex.com/admin/io-manager/explode/?forceLogs=true to see an error inside the iframe, with no request id

and 

https://anita--basedevmkp.myvtex.com/admin for full page error


beatrizmilhomem (changes requested):
Vendo agora com o refresh automático funcionando, deu pra ver que todo o grupo de elementos que vai da img até o botão dá um "pulo" pra baixo caso o refresh aconteça e o erro persista. Teria como manter na mesma posição? Sempre centralizado na página. Exemplo ilustrativo abaixo:

image

react/error.tsx
```diff
@@ -62,84 +53,9 @@ class ErrorPage extends Component {
...
-          <img
-            src={ErrorImg}
-            className={`${style.imgHeight} pb6 pb0-ns`}
-            alt=""
-          />
-        </div>
-      </div>
-    )
-  }
+    if (window.__ERROR__) console.error(window.__ERROR__)
```
lucasaarcoverde, line 56: Do you know if we can capture this error using Sentry? IMO we should avoid using console logs in prod

react/components/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import './shoreline.css'
+
+import styles from './error-page.module.css'
+import ErrorIcon from './ErrorIcon'
+import RefreshCounter from './RefreshCounter'
+
+export default function ErrorPage({ errorCode }: { errorCode?: string }) {
+  const date = new Date()
+  console.log('versao 2')
```
lucasaarcoverde, line 10:
```suggestion
```

react/components/ErrorPage/shoreline.css
marcelovicentegc, line 1: Do we need all these tokens?

react/components/ErrorPage/RefreshCounter.tsx
```diff
@@ -0,0 +1,49 @@
...
+      const timer = setInterval(() => setCounter(counter - 1), 1000)
+      return () => clearInterval(timer)
+    } else if (shouldRefresh) {
+      localStorage.setItem('hasRefreshed', 'true')
+      localStorage.setItem(
+        'lastRefreshTime',
+        (currentTime as unknown) as string
+      )
+
+      console.log('there was reload')
```
marcelovicentegc, line 29:
```suggestion
```

react/components/ErrorPage/RefreshCounter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    if (counter > 0) {
+      const timer = setInterval(() => setCounter(counter - 1), 1000)
+      return () => clearInterval(timer)
+    } else if (shouldRefresh) {
+      localStorage.setItem('hasRefreshed', 'true')
+      localStorage.setItem(
+        'lastRefreshTime',
+        (currentTime as unknown) as string
+      )
```
marcelovicentegc, line 27:
```suggestion
  const currentTime = new Date().toISOString()

  const shouldRefresh =
    !hasRefreshed ||
    (lastRefreshTime && currentTime - lastRefreshTime > EXPIRATION_TIME)

  useEffect(() => {
    if (counter > 0) {
      const timer = setInterval(() => setCounter(counter - 1), 1000)
      return () => clearInterval(timer)
    } else if (shouldRefresh) {
      localStorage.setItem('hasRefreshed', 'true')
      localStorage.setItem('lastRefreshTime', currentTime)
```

react/components/ErrorPage/RefreshCounter.tsx
marcelovicentegc, line 1: @anitavincent I miss the presence of a "clean up" logic. As it is now, once this screen is rendered to the user and a retry is executed, `hasRefreshed` is set to `true` and this breaks the `shouldRefresh` boolean if the user faces the error screen multiple times. It occurs to me that we might not even need the `hasRefreshed` variable in order to achieve what we want. It appears to me that the `lastRefreshTime` alone might do this job well

marcelovicentegc (commented):
Left a few suggestions :wink: 

We must have in mind that we'll need to export the ErrorIcon, RefreshCounter, error-page.module.css and shoreline.css from Render Server as well in order for this to work as expected in production :pray:

react/error.tsx
marcelovicentegc: Yes! We can! In fact, I think we already do :thinking:

marcelovicentegc: Yep, we do! Take a look at lines 34-52

image

anitavincent: I think we discuss it further, currently we show this error directly on the UI, since we won't be doing that any longer this was the proposal to simplify debug on error (nowadays people often just print the screen and share it for debug)

react/components/ErrorPage/RefreshCounter.tsx
anitavincent: Tested your suggestion, it didn't work. Because I need the date to be a number for this operation

` const shouldRefresh =
    !hasRefreshed ||
    (lastRefreshTime && currentTime - lastRefreshTime > EXPIRATION_TIME)` 
    
    using `new Date().toISOString()` didn't work for that operation, not even forcing it to be a number (forcing returns NaN)

react/components/ErrorPage/RefreshCounter.tsx
anitavincent: The current logic works, the refresh works again after 10 minutes. But you're right, the boolean hasRefreshed is useless, I removed it

marcelovicentegc: Perfect @anitavincent ! On point! :dart:

marcelovicentegc (commented):
@anitavincent , I still see :point_down: 

image

on https://anita--storecomponents.myvtex.com/admin/io-manager/explode/?forceLogs=true

anitavincent:
> @anitavincent , I still see 👇
> 
> image
> 
> on https://anita--storecomponents.myvtex.com/admin/io-manager/explode/?forceLogs=true

Oh that's because the error page is supposed to only appear in prod workspaces. Linked a version in this workspace now that forces the production version. So it should work

marcelovicentegc (approved):

beatrizmilhomem (approved):

vsseixaso (approved):

PR #669, opened 2024-11-11
feat: introduce i18n to error page

anitavincent:
#### What does this PR do? \*

introduces i18n to error page

#### How to test it? \*

To check full page error: https://anita--teamadmin.myvtex.com/admin

If your locale is pt-br some strings should be in portuguese (but not all)

for client side error:
https://anita1--teamadmin.myvtex.com/admin/io-manager/explode/?forceLogs=true


react/components/ErrorPage/messages/bg.json
```diff
@@ -0,0 +1 @@
+{}
```
lucasaarcoverde, line 1: shouldn't we need those translations before merging the PR?

lucasaarcoverde (approved):
LGTM, just a question

marcelovicentegc (commented):
Folks, I'm concerned with one aspect of these changes: a lot of times Render breaks is when it's going through its lifecycle that, among other things, **injects the locale** in the context.

This means that on real-world scenarios, we might just render the title IDs on the page with this approach in case the language is not available. This is why most of these strings are hardcoded and do not rely on react-intl.

react/components/ErrorPage/messages/bg.json
anitavincent: For now it just falls back to english so nothings is changed

anitavincent:
> Folks, I'm concerned with one aspect of these changes: a lot of times Render breaks is when it's going through its lifecycle that, among other things, **injects the locale** in the context.
> 
> This means that on real-world scenarios, we might just render the title IDs on the page with this approach in case the language is not available. This is why most of these strings are hardcoded and do not rely on react-intl.

There is a function that uses the english strings as fallback, the function could also handle an undefined locale and fall back to english. Isn't that enough?

sheilagomes (approved):

marcelovicentegc:
> > Folks, I'm concerned with one aspect of these changes: a lot of times Render breaks is when it's going through its lifecycle that, among other things, **injects the locale** in the context.
> > This means that on real-world scenarios, we might just render the title IDs on the page with this approach in case the language is not available. This is why most of these strings are hardcoded and do not rely on react-intl.
> 
> There is a function that uses the english strings as fallback, the function could also handle an undefined locale and fall back to english. Isn't that enough?

If we do have fallback, yes, we're good :wink:

react/components/ErrorPage/messages/index.ts
iago1501, line 1: Why aren't we using [messages builder](https://developers.vtex.com/docs/guides/vtex-io-documentation-messages-builder) instead?

react/components/ErrorPage/messages/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  'nl-NL': nl,
+  'ro-RO': ro,
+  'bg-BG': bg,
+  'th-TH': th,
+  'de-DE': de,
+}
+
+export function getMessages(locale: string) {
+  const localeMessages = (locale && messages[locale]) || {}
+  const fallbackMessages = messages['en-US']
```
marcelovicentegc, line 31: This is key and it answers @iago1501's question :heart:

iago1501: If it's because of the fallback we could simply use the available languages in vtex.store as an example of languages that we should take care in stores, for example: https://github.com/vtex-apps/store/tree/master/messages

or talking about admin, the ones that we have in admin-pages: https://github.com/vtex-apps/admin-pages/tree/master/messages

My main concern is that using the message's builder is one of our patterns when talking about translations, we have the messages folder and all translations are inside it, with this approach, talking about scaling, if any other component needs translation, we would have to add a messages folder for the component instead of center in a single place and this logic would be replied 🤔

marcelovicentegc: I see your point @iago1501, but given that this application didn't have translations until this day while being around for many years, I wouldn't be concerned about adopting a pattern that we might not need by the end of the day, and that we know it's the entry point for one of the most frequent errors we experience in the Admin given that the messages service itself didn't scale as we expected. 

For this same reason, it occurs to me that we should rather not rely on the messages service nor its builder, specially on a mission critical component such as this application, when we could otherwise organize translations where needed (the error component)  and skip all complexity that using the messages builder involves (triggering the messages service lifecycle just to fetch some strings stored elsewhere while it could just to be packaged with the application itself?).

It occurs to me that using the messages service could break the error component itself quite frequently.

gabpaladino:
@anitavincent acessei os links do PR mas só aparecem as msg misturadas (PT/EN), não tem um workspace com a solução funcionando?

anitavincent:
> @anitavincent acessei os links do PR mas só aparecem as msg misturadas (PT/EN), não tem um workspace com a solução funcionando?

oi! Quando eu lancei esses workspaces ainda não tinha todas as mensagens em portugues (por isso as mensagens misturadas) foi também um pouco pra testar o que acontecia no caso de umas mensagens estarem disponíveis. Observar o fallback em inglês funcionando. Nesses workspaces o locale tá só pt-br, um deles você consegue trocar o locale, o outro não porque o erro ocorre num nível acima antes de aparecer a UI pra isso.

O que você quer dizer com "solução funcionando". Posso re-subir esses workspaces com todas as strings em outras linguas que foram adicionadas desde então. Mas nesse caso o workspace com "erro global" que crasha tudo ainda vai mostrar só todo em português. É isso que você quer?

iago1501:
> > @anitavincent acessei os links do PR mas só aparecem as msg misturadas (PT/EN), não tem um workspace com a solução funcionando?
> 
> oi! Quando eu lancei esses workspaces ainda não tinha todas as mensagens em portugues (por isso as mensagens misturadas) foi também um pouco pra testar o que acontecia no caso de umas mensagens estarem disponíveis. Observar o fallback em inglês funcionando. Nesses workspaces o locale tá só pt-br, um deles você consegue trocar o locale, o outro não porque o erro ocorre num nível acima antes de aparecer a UI pra isso.
> 
> O que você quer dizer com "solução funcionando". Posso re-subir esses workspaces com todas as strings em outras linguas que foram adicionadas desde então. Mas nesse caso o workspace com "erro global" que crasha tudo ainda vai mostrar só todo em português. É isso que você quer?

É isso mesmo @anitavincent , é só pra poder ver funcional as traduções

iago1501 (approved):

react/components/ErrorPage/messages/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  'it-IT': it,
+  'nl-NL': nl,
+  'ro-RO': ro,
+  'bg-BG': bg,
+  'th-TH': th,
+  'de-DE': de,
+}
+
+export function getMessages(locale: string) {
+  const localeMessages = (locale && messages[locale]) || {}
```
iago1501, line 30:
```suggestion
  const fallbackMessages = messages['en-US']
  const localeMessages = (locale && messages[locale]) || fallbackMessages
  
  return localeMessages
```

Its just a suggestion, but it's up to you, to me it looks more readable this way

react/components/ErrorPage/messages/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  const fallbackMessages = messages['en-US']
+
+  return { ...fallbackMessages, ...localeMessages }
```
iago1501, line 33:
```suggestion
  const fallbackMessages = messages['en-US']
  const localeMessages = (locale && messages[locale]) || fallbackMessages

  return localeMessages
```

Just a suggestion, but it's up to you, to me this way looks more easy to read

react/components/ErrorPage/messages/index.ts
iago1501: Forgot to add the right lines 🥲

react/components/ErrorPage/messages/index.ts
iago1501: TL;DR;

We are worried about the usage of messages because most of the error screens are related to a failure in this service

react/components/ErrorPage/messages/index.ts
anitavincent: Sounds good I'll change it

anitavincent: Actually thinking a bit more, that destructing I did on line 33 was actually intentional, to make sure that even if we DO have messages (for examples for french) but only SOME keys are missing. So in this case of incomplete messages there would be no risk of some messages showing up as "render/error.message-name", the missing messages would show up in english instead

iago1501: Nowadays all messages are there for all mapped languages, right? I'm not sure if this is a real problem that we should worry about 🤔

anitavincent: I feel like it's safer to have the fallback still. More messages could be added with time, and I've seen it happen a few times if there are new feature in this page. Situations like clients seeing messages like "render/error.message-name" in production.

PR #674, opened 2025-03-26
[SFOC-95] Feat: add featureFlag to force native browser navigation

gabpaladino:
#### What does this PR do? \*

Revert to browser native navigation based on the feature flag `useDefaultBrowserNavigation`, this is causing a behavior in FastStore accounts where a user cannot navigate to different pages without fully reload

https://github.com/vtex-apps/store/pull/597

### DIFF
--- a/react/components/Link.tsx
+++ b/react/components/Link.tsx
@@ -70,10 +70,13 @@ const Link: React.FunctionComponent<Props> = ({
     rootPath = '',
     route: { domain },
     query: queryFromRuntime,
+    getSettings,
   } = useRuntime()
 
   // If workspace is set via querystring, keep it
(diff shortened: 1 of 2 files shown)


react/components/Link.tsx
```diff
@@ -122,7 +126,14 @@ const Link: React.FunctionComponent<Props> = ({
         event.preventDefault()
       }
     },
-    [to, onClick, navigate, target, options]
+    [
+      to,
+      storeSettings.useDefaultBrowserNavigation,
```
gabpaladino, line 131: curioso essa alteração nao fiz, foi automatica?! rs

iago1501 (changes requested):
- Changelog is missing, once it's done it's approved for deploy

PR #685, opened 2026-07-31
fix: load minified regenerator runtime in production

MarlonPassos-git:
#### What does this PR do? \*

Production storefronts load `regenerator-runtime@0.11.1/runtime.js` through the VTEX npm asset proxy. Lighthouse flags the 6.6 KiB transferred script as unminified JavaScript and estimates about 3.8 KiB savings. The same asset and audit were reproduced across five Store Framework storefronts.

Following [Iago's suggestion](https://github.com/vtex-apps/render-runtime/pull/685#issuecomment-5543945862), this PR now adds a local `react/regenerator-runtime.ts` entrypoint importing `regenerator-runtime/runtime`, pins the dependency to `0.11.1`, and points the production external at `runtime:regenerator-runtime`. Development keeps the existing npm `runtime.js` external.

The local entrypoint lets the VTEX React builder bundle and minify the runtime for production and serve it as a separate external from VTEX assets. This is expected to address Lighthouse's unminified JavaScript finding; the final gzip savings still need to be measured from the actual VTEX production output.

pagespeed-regenerator-runtime-audit

#### How to test it? \*

I could not link this app in my workspace; it appears that linking `vtex.render-runtime` requires VTEX access that I do not have. Validation of the new bundle is still pending and needs help from a maintainer with the required access.

1. Link the app in an authorized workspace or publish and install a beta version, then load a storefront using the production output.
2. Confirm that the `runtime:regenerator-runtime` asset loads before React/apps and exposes the `regeneratorRuntime` global in both the browser and SSR; check generator behavior and rendering for regressions.
3. Measure the production asset's gzip size against the existing approximately 6.8 KiB runtime, including the bundle wrapper, and rerun Lighthouse to verify the unminified JavaScript finding.
4. Confirm that development still loads the existing npm `runtime.js` external.

Superseded approach:

~~Change only the production external path to `runtime.min.js`. Although `regenerator-runtime@0.11.1` does not publish that file, jsDelivr generates missing `.min.js` variants on demand, and the VTEX npm asset proxy serves it through its jsDelivr fallback.~~

The PR now uses the locally imported, pinned runtime through the VTEX builder, as suggested by Iago, removing the dependency on jsDelivr's on-demand minification.

The previous measurements below apply **only to the superseded jsDelivr approach**, not to the new VTEX bundle:

- Existing runtime: 24,200 raw / about 6,833 gzip bytes.
- ~~jsDelivr minified variant: 6,674 raw / about 2,617 gzip bytes; about 4.2 KiB transferred savings.~~
- ~~The original and jsDelivr-minified responses exposed the same eight regenerator APIs and passed the same generator behavior probe.~~

[Review suggestion and pending validation](https://github.com/vtex-apps/render-runtime/pull/685#issuecomment-5543945862).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Update DK Catalog platform-flow-id
 
+### Fixed
+
+- Load the minified regenerator runtime in production
+
 ## [8.136.2] - 2026-03-02
 
 ### Added
--- a/react/externals.json
+++ b/react/externals.json
@@ -98,7 +98,7 @@
         "serverOnly": true
       },
       {
-        "path": "npm:regenerator-runtime@0.11.1/runtime.min.js"
+        "path": "runtime:regenerator-runtime"
       },
       {
         "global": "PropTypes",
--- a/react/externals.json
+++ b/react/externals.json
@@ -98,7 +98,7 @@
         "serverOnly": true
       },
       {
-        "path": "npm:regenerator-runtime@0.11.1/runtime.js"
+        "path": "npm:regenerator-runtime@0.11.1/runtime.min.js"
       },
       {
         "global": "PropTypes",
--- a/react/package.json
+++ b/react/package.json
@@ -33,6 +33,7 @@
     "react-apollo": "^3.1.2",
     "react-helmet": "^5.2.0",
     "react-json-view": "^1.19.1",
+    "regenerator-runtime": "0.11.1",
     "route-parser": "^0.0.5",
     "use-media": "^1.4.0"
   },
@@ -78,7 +79,6 @@
     "react-hot-loader": "^4.3.12",
     "react-intl": "^3.9.1",
     "react-no-ssr": "^1.1.0",
-    "regenerator-runtime": "^0.13.1",
     "ts-jest": "^24.0.1",
     "typescript": "3.9.7"
   },
--- a/react/regenerator-runtime.ts
+++ b/react/regenerator-runtime.ts
@@ -1,0 +1,1 @@
+import 'regenerator-runtime/runtime'
--- a/react/yarn.lock
+++ b/react/yarn.lock
@@ -7229,7 +7229,7 @@ regenerate@^1.2.1, regenerate@^1.4.0:
   resolved "https://registry.yarnpkg.com/regenerate/-/regenerate-1.4.0.tgz#4a856ec4b56e4077c557589cae85e7a4c8869a11"
   integrity sha512-1G6jJVDWrt0rK99kBjvEtziZNCICAuvIPkSiUFIQxVP06RCVpq3dmDo2oi6ABpYaDYaTRr67BEhL8r1wgEZZKg==
 
-regenerator-runtime@^0.11.0:
+regenerator-runtime@0.11.1, regenerator-runtime@^0.11.0:
   version "0.11.1"
   resolved "https://registry.yarnpkg.com/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz#be05ad7f9bf7d22e056f9726cee5017fbf19e2e9"
   integrity sha512-MguG95oij0fC3QV3URf4V2SDYGJhJnJGqvIIgdECeODCT98wSWDAJ94SSuVpYQUoTcGUIL6L4yNB7j1DFFHSBg==


MarlonPassos-git:
vtexio/version-select failed because **Patch** is not selected in the VTEX IO CI/CD bot comment. As an external contributor, I cannot mark that option or access the CI details, so a VTEX maintainer needs to select **Patch** and rerun the CI.

iago1501:
Hi @MarlonPassos-git ... Thanks for digging into this and for measuring it across storefronts. The diagnosis is spot on: production loads unminified regenerator-runtime@0.11.1/runtime.js, and Lighthouse is right to flag it.

We looked at this with the team and our suggestion is to take a slightly different path, if you are up for it.

runtime.min.js is not published on npm (unpkg 404s; later versions do not ship it either), so the asset proxy has to lean on jsDelivr’s on-demand minify as the only working origin. That file is a critical runtime for both the browser and SSR, so we would rather not make a third-party fallback the source of truth for it.

The savings you measured (~4 KiB gzip) are real and appreciated. A path that should keep that win without the extra dependency:

Add a react/ entrypoint (e.g. regenerator-runtime) pinned to 0.11.1, same idea as runtime:AMP / runtime:intl-polyfill.
Point prod externals.json at runtime:regenerator-runtime, and keep dev on runtime.js.

The react builder already minifies prod output, the script stays a separate external (load order before React/apps stays the same), and origin/cache stay on VTEX published assets.

Two things worth checking if you go that way: the webpack bundle still has to expose the regeneratorRuntime global on client and SSR, and the gzip size should still beat today’s ~6.8 KiB. If the chunk wrapper eats the savings, it is probably not worth the extra moving parts.

Happy to help review that follow-up if you want to try it. Thanks again for the PR, this was a useful catch.

MarlonPassos-git:
Hi @iago1501, how are you? Thanks for the feedback — using a local entrypoint was a great suggestion!

I've updated the PR to import `regenerator-runtime/runtime` locally, pinned to `0.11.1`, and use `runtime:regenerator-runtime` in production, keeping development on the existing `runtime.js` external.

Unfortunately, I couldn't link the app in my workspace. It appears that linking `vtex.render-runtime` requires VTEX access that I don't have. Would you be able to help validate it in a workspace where you can link this app, or through a beta release?

To get the actual numbers, we still need to measure the production output from that build, including the bundle wrapper, and confirm that `regeneratorRuntime` is exposed in both the browser and SSR. Since the React builder minifies production output, this should address the unminified JavaScript finding, but the final gzip savings and Lighthouse result still need that validation.

Thanks again for the guidance!

iago1501:
Hi @MarlonPassos-git, great work here! 🎉 I set up a `production=true` workspace (`qareviewprod`) with your branch installed and validated the bundled-entrypoint approach end-to-end — and it not only works correctly, it beats the numbers we were hoping for.

### Results

| | Raw size | Transferred (gzip) |
|---|---:|---:|
| **Before** (`npm:regenerator-runtime@0.11.1/runtime.js`) | 24,124 B | 6,791 B |
| **After** (`runtime:regenerator-runtime`, your bundled entry) | 6,964 B | **2,574 B** |
| **Savings** | -71% | **-62% (−4,217 B per page load)** |

Re-confirmed independently on a second production domain (`storetheme.vtex.com?workspace=qareviewprod`, served from the public `vtexassets.com` CDN): same chunk, same behavior, **2,524 B** gzip — consistent with the numbers above.

### What I checked

- ✅ Only the new bundled chunk (`.../vtex.render-runtime@8.136.3-beta.2/public/react/regenerator-runtime.min.js`) loads — the old npm external is **not requested at all** anymore.
- ✅ `window.regeneratorRuntime` is fully functional (`wrap`, `mark`, `isGeneratorFunction`, `awrap`, `async`, all present as expected).
- ✅ Correctly reflected in the SSR HTML (only the new chunk's `<script>` tag appears).
- ✅ No console errors related to generators/async functions; GraphQL and other async-dependent calls all worked normally.
- ✅ Dev workspaces (confirmed via `vtex link`) correctly keep using the old npm path, exactly as intended by your externals.json split.

This beats the ~4 KiB savings estimate from the original (superseded) jsDelivr approach, while avoiding the third-party fallback dependency entirely. Both open questions from my earlier review — whether `regeneratorRuntime` stays exposed correctly and whether the final gzip size would still beat today's baseline — are answered: yes, on both counts, by a solid margin.

Thanks for iterating on this and for the clean, well-scoped fix. 🚀

iago1501 (approved):

MarlonPassos-git:
@iago1501 Now that everything is sorted out, do you think we could get this merged into production in the next few weeks?

PR #686, opened 2026-09-04
perf: lazy-load Sentry SDK via dynamic import (admin-only)

iago1501:
### 📉 -42% render-runtime JS on the storefront (~1.7 MB), 0 behavior change for the admin

| | Before | After |
|---|---|---|
| Storefront (6 entrypoints, prod account) | 4048.4 KB | **2348.4 KB** (**-1700 KB / -42%**) |
| Dev workspace (unminified) | 3992.9 KB | **2347.9 KB** (**-41%**) |
| Sentry requests on storefront | full SDK bundled in `common.js` | **0 bytes / 0 requests** |
| Sentry on `/admin/` | eager, same session | fetched on demand, same events still captured |

#### What does this PR do? *

`@sentry/react` (+ Replay) was statically imported even though it's only ever used behind `isAdmin()`. Static imports can't be tree-shaken, so **every** storefront visitor downloaded the full SDK for nothing.

This PR adds `react/o11y/sentry.ts`, a thin wrapper that lazily `import()`s `@sentry/react` into its own chunk (`sentry.js`), only from `isAdmin()`-gated call sites. `error.tsx`/`ErrorBoundary.tsx` now use its drop-in async `captureException()`; config (DSN, replay sampling, `beforeSend` tagging) is unchanged.

#### How to test it? *

1. Link the branch, load the storefront home page → no `sentry.js` request, no Sentry SDK in `common.js`.
2. Load `/admin/` → `sentry.js` fetched on demand, `window.__SENTRY__` initializes, errors are still captured (verified a real event reaching Sentry's ingest endpoint).

No new console errors, no FCP/LCP/DCL regression in either environment.

### DIFF
--- a/react/error.tsx
+++ b/react/error.tsx
@@ -1,6 +1,6 @@
 /* global module */
 import React, { Component } from 'react'
-import { captureException } from '@sentry/react'
+import { captureException } from './o11y/sentry'
 
 require('myvtex-sse')
 
--- a/react/o11y/instrument.ts
+++ b/react/o11y/instrument.ts
@@ -1,60 +1,9 @@
-import * as Sentry from '@sentry/react'
 import { isAdmin } from '../utils/isAdmin'
-import { getIOContext } from './ctx'
+import { initSentry } from './sentry'
 
+// Sentry is only ever used in the admin. Loading it lazily (dynamic import)
+// keeps it out of the storefront's main bundle entirely, since bundlers can
+// only tree-shake/split code that isn't statically imported.
 if (isAdmin()) {
-  Sentry.init({
-    dsn:
-      'https://2fac72ea180d48ae9bf1dbb3104b4000@o191317.ingest.us.sentry.io/1292015',
-    integrations: [Sentry.replayIntegration()],
-
-    // Set tracesSampleRate to 0.1 to capture 10%
-    // of transactions for tracing.
-    tracesSampleRate: 0.1,
-
-    // Set `tracePropagationTargets` to control for which URLs trace propagation should be enabled
-    tracePropagationTargets: [
-      /^\//,
-      /^(https?:\/\/)?([a-z0-9]+[.])*myvtex\.com/,
-    ],
-
-    // Capture Replay for 0% of all sessions,
-    // plus for 50% of sessions with an error
-    replaysSessionSampleRate: 0,
-    replaysOnErrorSampleRate: 0.5,
-
-    beforeSend: (event) => {
-      const ctx = getIOContext()
-
-      // Must check with false, otherwise default null's
-      // value leads to data mistakenly not sent to Sentry,
-      // which can occur if somehow we can't infer whether
-      // the apps are running under a production or development
-      // environment.
-      if (ctx.admin_production === false) {
-        const params = new URL(document?.location?.toString())?.searchParams
-        const shouldLog = params.get('forceLogs')
-
-        if (shouldLog === 'true') {
-          return makeEventWithCtx(event, ctx)
-        }
-
-        return null
-      }
-
-      return makeEventWithCtx(event, ctx)
-    },
-  })
-}
-
-function makeEventWithCtx(event: any, ctx: any) {
-  const eventWithCtx = {
-    ...event,
-    tags: {
-      ...event?.tags,
-      ...ctx,
-    },
-  }
-
-  return eventWithCtx
+  initSentry()
 }
--- a/react/o11y/instrument.ts
+++ b/react/o11y/instrument.ts
@@ -1,9 +1,13 @@
-import { isAdmin } from '../utils/isAdmin'
-import { initSentry } from './sentry'
-
-// Sentry is only ever used in the admin. Loading it lazily (dynamic import)
-// keeps it out of the storefront's main bundle entirely, since bundlers can
-// only tree-shake/split code that isn't statically imported.
-if (isAdmin()) {
-  initSentry()
-}
+// Sentry is only ever used in the admin, and only to report errors. Loading
+// it lazily (dynamic import) keeps it out of the storefront's main bundle
+// entirely, since bundlers can only tree-shake/split code that isn't
+// statically imported.
+//
+// This file intentionally does NOT eagerly call `initSentry()`: doing so
+// would make every admin page pay the extra `sentry.js` chunk
+// request/parse/exec cost on every session, even when no error ever
+// occurs, which defeats the "on demand" goal. Instead, the SDK is loaded
+// and initialized lazily, the first time `captureException`/`initSentry`
+// is actually invoked from an admin-gated error path (see
+// react/error.tsx and react/components/ErrorBoundary.tsx).
+export {}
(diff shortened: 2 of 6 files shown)


react/o11y/sentry.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  return eventWithCtx
+}
+
+/**
+ * Lazily loads the Sentry SDK. It is only ever invoked from admin-gated
+ * call sites (see isAdmin() checks in error.tsx/ErrorBoundary.tsx), so
+ * storefront (non-admin) bundles never pay for the dynamic chunk.
+ */
+function loadSentry(): Promise<SentryModule> {
```
mendescamara, line 32: `loadSentry()` caches the `import()` promise in `sentryModulePromise` and never resets it on rejection. If the dynamic import fails once during a page session (e.g. an ad-blocker blocking a chunk literally named `sentry.js`, or a transient network error), `sentryModulePromise` stays pointed at a rejected promise forever — every subsequent `initSentry()`/`captureException()` call in that session immediately re-rejects instead of retrying, silently disabling Sentry for the rest of the session. This failure mode didn't exist with the previous static import.

Consider resetting `sentryModulePromise` (and `initialized`) back to `null`/`false` in a `.catch()` so a later call can retry.

react/error.tsx
```diff
@@ -1,6 +1,6 @@
 /* global module */
 import React, { Component } from 'react'
-import { captureException } from '@sentry/react'
+import { captureException } from './o11y/sentry'
```
mendescamara, line 3: None of the call sites await or `.catch()` the new async `captureException`/`initSentry` (this import is the same pattern used at `ErrorBoundary.tsx:32` and `instrument.ts`'s top-level `initSentry()` call). Since `captureException` used to be a synchronous, non-throwing export of `@sentry/react`, the `try/catch` in `componentDidMount` below (lines 37-52) can no longer catch anything it rejects with — a rejection (e.g. from the caching issue on `loadSentry()`) now surfaces only as an unhandled promise rejection in the console instead of being handled.

Worth adding a `.catch(() => {})` (or a shared no-op handler) at each call site.

react/o11y/sentry.ts
```diff
@@ -0,0 +1,105 @@
+import { isAdmin } from '../utils/isAdmin'
+import { getIOContext } from './ctx'
+
+type SentryModule = typeof import('@sentry/react')
+
+let sentryModulePromise: Promise<SentryModule> | null = null
```
mendescamara, line 6: This module introduces new stateful logic (import caching via `sentryModulePromise`, one-time init via `initialized`) with no unit tests. A test exercising the reject/retry path would likely have caught the caching issue flagged above — worth adding coverage before merging.

mendescamara (commented):
Reviewed the lazy-loading approach — the goal and measured bundle-size win are solid. Left 3 inline notes on the new `sentry.ts` wrapper around failure handling.

react/o11y/sentry.ts
iago1501: Fixed in 2e7217e: `loadSentry()` now resets `sentryModulePromise` back to `null` in a `.catch()` when the dynamic import rejects (and `ensureInitialized()` resets `initialized` if `Sentry.init` throws), so a transient chunk-load failure no longer disables Sentry for the rest of the session — the next call retries. Also added test coverage for this exact reject-then-retry path in 7283a06 (`react/o11y/sentry.test.ts`).

react/error.tsx
iago1501: Fixed in 2e7217e: `initSentry()`/`captureException()` now catch any failure internally (chunk-load error, `Sentry.init` throwing, etc.) and log it via `console.error` instead of letting the promise reject, since every call site still invokes them fire-and-forget. So a failure here never surfaces as an unhandled rejection anymore, and (importantly) it no longer masks/interferes with the caller's own error handling below.

react/o11y/sentry.ts
iago1501: Added in 7283a06 (`react/o11y/sentry.test.ts`): coverage for the caching/init state, including the exact reject-then-retry scenario you flagged (a chunk-load failure followed by a later successful call), plus the admin-gating no-op paths and the 'initialize once across multiple calls' behavior.

react/o11y/sentry.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  try {
+    doInit(Sentry)
+  } catch (error) {
+    // Allow a later call to retry `Sentry.init` instead of permanently
+    // treating the SDK as initialized when it actually threw.
+    initialized = false
+    throw error
+  }
+}
```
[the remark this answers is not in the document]
iago1501: Thanks for flagging, but I don't think this is racy in practice: `doInit(Sentry)` calls `Sentry.init(...)` synchronously (no `await` inside it), and there's no `await`/yield point between `initialized = true` and `doInit()` completing inside `ensureInitialized`. Since JS runs to the next `await`/yield before any other queued call can resume, a second concurrent call can only observe `initialized === true` *after* `doInit` has already fully run — there's no window where the flag is `true` but init is still in flight. Verified this holds in the Babel-transpiled (regenerator) output used by the test suite too (no yield before `doInit`). If `doInit`/`Sentry.init` ever becomes genuinely async in the future, this would need revisiting (e.g. guarding with an in-flight promise as suggested) — happy to add that defensively if you'd prefer not to rely on this invariant.

react/o11y/instrument.ts
```diff
@@ -1,60 +1,13 @@
...
-  }
-
-  return eventWithCtx
-}
+// Sentry is only ever used in the admin, and only to report errors. Loading
+// it lazily (dynamic import) keeps it out of the storefront's main bundle
+// entirely, since bundlers can only tree-shake/split code that isn't
+// statically imported.
+//
+// This file intentionally does NOT eagerly call `initSentry()`: doing so
```
mendescamara: This changes Admin observability behavior, even though the PR states there is no behavior change for Admin. With `replaysSessionSampleRate: 0` and `replaysOnErrorSampleRate: 0.5`, Replay needs Sentry to be initialized before the error so it can continuously keep the pre-error buffer (up to 60 seconds). Initializing only inside `captureException()` means that context is already lost; in Sentry v7 the Replay integration also starts asynchronously after `init`, so the triggering exception may not get a Replay at all. Breadcrumbs and automatic global error handling are likewise unavailable before the first explicit capture. Could we keep the dynamic import but call fire-and-forget `initSentry()` here during Admin startup? Its internal `isAdmin()` guard would preserve zero Sentry chunk requests on storefront while retaining the previous Admin behavior.

marcelovicentegc (approved):

react/o11y/instrument.ts
iago1501: Good catch, you're right — fixed in d58d293. `instrument.ts` now calls `initSentry()` eagerly again (fire-and-forget), same as before the lazy-load work, so the admin keeps the pre-error Replay buffer, breadcrumbs, and Sentry's own global `window.onerror`/`onunhandledrejection` handlers active from page load. `initSentry()` still no-ops outside `isAdmin()`, so the storefront still never fetches the `sentry.js` chunk — the only thing reverted is the admin's own "fetch on demand vs. fetch on every session" trade-off, in favor of keeping full observability there, as originally intended by this PR.

mendescamara (approved):

PR #688, opened 2026-09-10
Revert "fix: load minified regenerator runtime in production" (#685)

iago1501:
#### What does this PR do? *

Reverts #685 ("fix: load minified regenerator runtime in production", shipped in `v8.136.4`).

That PR moved the production `regenerator-runtime` external from a plain npm path to `runtime:regenerator-runtime`, a new local bundler entrypoint. Every entrypoint participates in webpack's `splitChunks`, so this new entrypoint pulled in the shared `common` chunk as a dependency. The HTML emitter ended up shipping two `async` `common.min.js` `<script>` tags on every page — one from the main `render-runtime` bundle, one from the new `regenerator-runtime` entrypoint.

The duplicated `async` scripts raced during client-side hydration. On heavier storefronts this broke header/navigation rendering intermittently. Prometheus (`order_transaction_begin_per_account_total`) shows a 50–70% drop in orders for ~1.5h on the most affected accounts (`americanino`, `lennyniemeyer`, `usaflextatix`) right after the `8.136.4` rollout, related to [#inc-6347-esc](https://github.com/vtex-apps/render-runtime/issues). Rolling the affected accounts back to `8.136.3` resolved it, confirming `8.136.4` as the regression.

This PR:
- Reverts `react/externals.json`, `react/regenerator-runtime.ts`, `react/package.json`, `react/yarn.lock` back to the pre-#685 state (plain `npm:regenerator-runtime@0.11.1/runtime.js` external, no local entrypoint).
- Adds a CHANGELOG entry under `[Unreleased]` documenting the revert and why.
- Marks `## [8.136.4]` as `[YANKED]` in the CHANGELOG, since that published version contains the regression and should not be installed.

#### How to test it? *

1. Confirm `react/externals.json`'s `prod.commons` block for `regenerator-runtime` is back to `npm:regenerator-runtime@0.11.1/runtime.js` (matches `dev`).
2. Build/link on a `production=true` workspace and confirm the rendered HTML has exactly **one** `common.min.js` `<script>` tag.
3. Confirm `regeneratorRuntime` is still available as a global (generator/async-await behavior unaffected).

#### Describe alternatives you've considered, if any. *

Keeping #685's minification goal but implementing it safely (e.g. importing `regenerator-runtime/runtime` directly inside `react/index.tsx` instead of a separate bundler entrypoint, so it never becomes its own chunk-splitting root) is a good follow-up, but out of scope for this revert — this PR only restores the known-good state so `master` is unblocked.

#### Related to / Depends on *

Reverts #685. Related to #inc-6347-esc.


mendescamara (approved):

MarlonPassos-git:
Hi @iago1501, I opened #690 with the safer version you suggested: the pinned runtime is now imported inside the `react/index.tsx` entrypoint instead of a new client entrypoint, and the npm external is removed. Same ~4 KiB savings target without the shared-chunk duplication that caused inc-6347-esc.

I still can't link `vtex.render-runtime` in my workspace, so if you can validate it again in a `production=true` workspace (the checklist is in the PR description), it would really help. Thanks!

